PR #7, opened 2019-11-03
added four different syntax highlight/colours

yCobanoglu:
- Editor React Component now Class (instead of functions)
- Syntax Highlighting
- |asdasdjh | and #asdasd have somewhat close colors
- AtfMode.js has to be typescript ?


src/editor/Editor.tsx
```diff
@@ -27,38 +28,62 @@ const specialCharacterKeys = Object.entries(specialCharacters).map(
...
 )
+interface Props {
+  readonly name: string
+  readonly value: string
+  readonly onChange: (value: string, event?: any) => void | undefined
+  readonly disabled: boolean
+  readonly error: object
+}
+
+class Editor extends Component<Props> {
```
jlaasonen: Since the introduction of [hooks ](https://reactjs.org/docs/hooks-intro.html) refs etc. can also be used in function components.

src/editor/Editor.tsx
```diff
@@ -27,38 +28,62 @@ const specialCharacterKeys = Object.entries(specialCharacters).map(
...
-      // @ts-ignore
-      setOptions={{
-        showLineNumbers: false,
-        newLineMode: 'unix'
-      }}
-      // @ts-ignore
-      commands={specialCharacterKeys}
-    />
-  )
+  render() {
```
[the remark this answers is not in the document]
jlaasonen: The length is due to many props in the Ace editor component.

src/editor/AtfMode.js
```diff
@@ -0,0 +1,40 @@
...
+          token: 'constant.numeric',
+          regex: '^#.*$'
+        },
+        {
+          token: 'string',
+          regex: '\\|[^\\|]*\\|'
+        },
+        {
+          token: 'support.function',
+          regex: '\\[[^\\[]*\\]'
```
jlaasonen: The regex does not match the description.

src/editor/AtfMode.js
```diff
@@ -0,0 +1,40 @@
...
+        {
+          token: 'constant.numeric',
+          regex: '^#.*$'
+        },
+        {
+          token: 'string',
+          regex: '\\|[^\\|]*\\|'
+        },
+        {
+          token: 'support.function',
```
jlaasonen: The colour is very close to the text colour in the current theme. 

The token types seem to be semantically wrong. Unfortunately, Ace was meant to be a code editor and tokens match that domain. While it is impossible to get 100% "correct" tokens we should try to find meaningful values. See: [Common Tokens](https://github.com/ajaxorg/ace/wiki/Creating-or-Extending-an-Edit-Mode#commonTokens). Ideally, we should create our own theme, but that's probably worth its own card.

Suggestions for tokens:

- **#**: `comment.line.number`
- **@**: `keyword` or `meta`. These lines mark on which side, column etc. the text is on.
- **$**: `markup`. These lines mark underlinings, blank lines etc.

PR #11, opened 2020-03-20
Display mode for $

yCobanoglu:


yCobanoglu:
[Display mode for Lines beginning with RulingDollarLine](https://trello.com/c/CHRE58CU/350-display-mode-for-lines-beginning-with-rulingdollarline)

src/fragmentarium/ui/display/Transliteration.tsx
```diff
@@ -143,16 +151,60 @@ function DisplayLine({
...
+          className={`Transliteration__${line.type} ${rulingDollarLine.number}`}
+        />
+      ]
+    )
+  } else {
+    return React.createElement(
+      container,
+      { className: classNames([`Transliteration__ControlLine`]) },
+      [
+        <div key={101} className="Transliteration__DollarLineAndAtLine">
```
yCobanoglu: temporary solution to pass a random prop.

src/fragmentarium/ui/display/Transliteration.tsx
```diff
@@ -143,16 +151,60 @@ function DisplayLine({
...
+        <hr
+          key={101}
+          className={`Transliteration__${line.type} ${rulingDollarLine.number}`}
+        />
+      ]
+    )
+  } else {
+    return React.createElement(
+      container,
+      { className: classNames([`Transliteration__ControlLine`]) },
```
yCobanoglu: My first choice was to pass `null` as props. The current `className` passed as Props isn't used by CSS

src/fragmentarium/ui/display/Transliteration.tsx
jlaasonen: That is not a random key. Its a constant.

src/fragmentarium/domain/text.ts
```diff
@@ -140,16 +140,23 @@ export type Token =
   | Enclosure
   | Tabulation
 
+export type LineContainer = Line | RulingDollarLine
```
jlaasonen: This is not a container. It also seems redundant since `RulingDollarLine` is a `Line`.

How about renaming the `Line` below and having this like the tokens above `Line = RulingDollarline | ...`?

src/fragmentarium/ui/display/Display.sass
```diff
@@ -14,23 +14,46 @@
...
 @mixin emesal
-  &--EMESAL,
+  &--EMESAL
     @include normal-font
     color: pink
     letter-spacing: 0.1em
 
+
+@mixin third-black-border
+  box-shadow: 0 1.3px white,0 2.6px black
```
jlaasonen: This is very confusing, it is not a border and not third. Why there is even need for a mixin?

src/fragmentarium/ui/display/Display.sass
```diff
@@ -14,23 +14,46 @@
...
+      margin-bottom: 6px
+
+    &.DOUBLE
+      border-bottom: 4px double black
+      margin-bottom: 9px
+
+    &.TRIPLE
+      border-bottom: 4px double black
+      margin-bottom: 15px
+      @include third-black-border
```
jlaasonen: Use [BEM](http://getbem.com) naming conventions for CSS classes.

jlaasonen: Why not use `border-top`?

src/fragmentarium/ui/display/Transliteration.tsx
```diff
@@ -143,16 +151,60 @@ function DisplayLine({
...
+  container?: string
+}): JSX.Element {
+  if (line.type === 'RulingDollarLine') {
+    const rulingDollarLine = line as RulingDollarLine
+    return React.createElement(
+      container,
+      { className: classNames([`Transliteration__ControlLine`]) },
+      [
+        <hr
+          key={101}
```
jlaasonen: Why are you doing thins? Is it not possible to pass a single child?

src/fragmentarium/ui/display/Transliteration.tsx
```diff
@@ -143,16 +151,60 @@ function DisplayLine({
...
+  container = 'div'
+}: {
+  line: LineContainer
+  container?: string
+}): JSX.Element {
+  if (line.type === 'RulingDollarLine') {
+    const rulingDollarLine = line as RulingDollarLine
+    return React.createElement(
+      container,
+      { className: classNames([`Transliteration__ControlLine`]) },
```
jlaasonen: Please use the correct element name. This is `RulingDollarLine` not `ControlLine`.

src/fragmentarium/ui/display/Transliteration.tsx
jlaasonen: Please use the correct element name. This not `ControlLine`.

src/fragmentarium/ui/display/Transliteration.tsx
jlaasonen: Why not just pass a single child?

Why have to elements with element class? Use the correct element class above and just use plain string or span here.

src/fragmentarium/ui/display/Transliteration.tsx
```diff
@@ -143,16 +151,60 @@ function DisplayLine({
...
+        />
+      ]
+    )
+  } else {
+    return React.createElement(
+      container,
+      { className: classNames([`Transliteration__ControlLine`]) },
+      [
+        <div key={101} className="Transliteration__DollarLineAndAtLine">
+          ({line.content[0].value.trimStart()})
```
jlaasonen: Do not use `content` it will be removed once all fragments are updated to the newest state. Dollar- and at-lines have the proper model you implemented earlier. If it seems to bee too complicated to generate the content in the fronted (e.g. state lines) consider adding a property for display value in the API.

src/fragmentarium/ui/display/Transliteration.tsx
```diff
@@ -143,16 +151,60 @@ function DisplayLine({
...
 }: {
   text: Text
 }): JSX.Element {
   return (
     <ol className="Transliteration">
-      {lines.map((line: Line, index: number) => (
-        <DisplayLine key={index} container="li" line={line} />
-      ))}
+      {lines.map((line: Line, index: number) => {
+        const LineComponent = lineComponents.get(line.prefix) || DisplayLine
```
jlaasonen: Do not use the `prefix` use `type`. The prefix will also be removed along with `content`.

src/fragmentarium/domain/text.ts
```diff
@@ -140,16 +140,23 @@ export type Token =
...
+
 export interface Line {
   readonly type: string
   readonly prefix: string
   readonly content: ReadonlyArray<Token>
 }
 
+export interface RulingDollarLine extends Line {
```
jlaasonen: `type` should be constant `RulingDollarLine`.

src/fragmentarium/ui/display/Transliteration.tsx
```diff
@@ -143,16 +151,60 @@ function DisplayLine({
...
 }
 
+function DollarLineAndAtLine({
+  line,
+  container = 'div'
+}: {
+  line: LineContainer
+  container?: string
+}): JSX.Element {
+  if (line.type === 'RulingDollarLine') {
```
jlaasonen: Why don't you make a separate component for the `RulingDollarLine`?

src/fragmentarium/ui/display/Transliteration.tsx
```diff
@@ -143,16 +151,60 @@ function DisplayLine({
   )
 }
 
+function DollarLineAndAtLine({
```
jlaasonen: It is not a good idea to make this kind of combo function. All the lines have a different model and you will end up with a huge if-else-block inside. You already have the logic below the select the correct component type.

If you find a common denominator and want to make a component based on that, name it accordingly.

jlaasonen: You split the card to ruling and others, so it does not make sense to care about other types of lines here.

src/fragmentarium/ui/display/Transliteration.tsx
```diff
@@ -143,16 +150,48 @@ function DisplayLine({
...
+  line,
+  container = 'div'
+}: {
+  line: Line
+  container?: string
+}): JSX.Element {
+  const rulingDollarLine = line as RulingDollarLine
+  const element = `Transliteration__${rulingDollarLine.type}`
+  const number = capitalizeFirstLetter(
+    rulingDollarLine.display_value.split(' ')[0]
```
jlaasonen: Use the correct property from the line model (`number`) instead of adding hacks.

Maybe adding the display value is not a good idea after all?

src/fragmentarium/domain/text.ts
```diff
@@ -140,11 +140,31 @@ export type Token =
...
+  prefix_modifier?: string
+  suffix_modifier?: string
+}
+export interface TextLine extends LineBase {
+  readonly line_number?: LineNumber
+}
+
+export interface ControlLines extends LineBase {
+  readonly display_value: string
+}
```
jlaasonen: You should concentrate on getting https://trello.com/c/CHRE58CU done and not add all the new stuff here. There is no point in splitting the card if you are going to put everything in one huge pull request again. We should use [isolated feature branches](https://www.martinfowler.com/bliki/FeatureBranch.html) and not stick all work in progress in a single branch. It is also faster and easier to review concise pull requests.

Also, the property names are wrong. JavaScript convention is to use camel-case property names. Please consult the schemas for correct names (there are some cases where snake-case is erroneously used). All new code should use proper naming conventions.

src/fragmentarium/ui/display/Transliteration.tsx
```diff
@@ -143,16 +150,48 @@ function DisplayLine({
...
+  )
+  return React.createElement(
+    container,
+    { className: classNames([element]) },
+    <div className={`${element}${number}`} />
```
jlaasonen: Use `hr` for rulings. 

Having two nested dollar line elements seems weird. What about using a modifiuer instead. E.g. `Transliteration__RulingDollarLine Transliteration__RulingDollarLine--SINGLE`.

yCobanoglu: So i used `.SINGLE`instead of modifiers. 
I wasn't able to create the line effect with `<hr/>' . Using `<hr/>`with *border-top & border-top* leads to the middle horizontal line (`<hr/>') not to be centered in the screen. I could alternatively use three `<hr/> within one `<div>`thats scaled.

src/fragmentarium/domain/text.ts
```diff
@@ -140,11 +140,31 @@ export type Token =
...
+
+export interface RulingDollarLine extends ControlLines {
+  readonly number: string
+  readonly status?: string | null
+}
```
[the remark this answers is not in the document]
jlaasonen: Use `"SINGLE" | "DOUBLE" | "TRIPLE" `.

src/fragmentarium/ui/display/Transliteration.tsx
jlaasonen: Why do you invent your conventions instead of using BEM?

Why do you need a `div` with `hr`?

What's the issue with horizontal centring? There should no difference between `hr` and `div. You know all the sizes so it should be a matter of adding correct margins.

jlaasonen: If you do not like to use modifiers, you could have elements `Transliteration__single-ruling` etc.

jlaasonen: That might be semantically more "correct" than letting ATF terminology leak into CSS.

yCobanoglu: Ok to use `Transliteration__single-ruling` i would have to manipulate the `displayValue`property by replacing the space with and - .
And for the `<hr>`tag.  I dont think its possible to create three horizontal lines with borders within `<hr>`.

jlaasonen: Do not use `displayValue`.

jlaasonen: Why not? It is a HTML-element like any other. It just has some default styles.

yCobanoglu: i could use a map
```javascript
const rulingNumber = new Map[
["SINGLE", "single-ruling"],
["DOUBLE", "single-ruling"],
["TRIPLE", "single-ruling"]
]
```
to create the css classnames

yCobanoglu: we will need `displayValue` for stateLine though right ?

jlaasonen: The same styles which work for the `div` should work for the `hr`.

yCobanoglu: ok. I will try it some more. And to create the css classnames i'll use a map and match the `number` property of RUlingDollarLine against this map ?

jlaasonen: You can use the map or convert `number` to lower case.

We do not need `displayValue` at all. All the required data is already there, but it might be easier to create that in the API, since we already have the logic for `content`. 

Display value makes no sense for the rulings since they are not displayed as text.

yCobanoglu: ok <hr> is now implemented

src/fragmentarium/ui/display/Display.sass
```diff
@@ -14,23 +14,39 @@
...
+    border-top: 2px solid black
+    height: 1px
+    margin: 0
+
+    &--double
+      border-bottom: 2px solid black
+      padding: 0.375px
+    &--triple
+      border-bottom: 6px double black
+      padding: 0.75px
```
jlaasonen, line 48: Could use only `height` (or only `padding` on a single side if there are issues with height) now it is hard to figure out the exact space between the borders.

yCobanoglu: yeah height will do it !

PR #14, opened 2020-04-08
Format brackets in glosses

jlaasonen:


jlaasonen:
[Format brackets in glosses](https://trello.com/c/OfZ641qX/357-format-brackets-in-glosses)

PR #15, opened 2020-04-09
thin rulings

yCobanoglu:


src/fragmentarium/ui/display/Display.sass
```diff
@@ -37,18 +37,17 @@
...
 
     &--double
-      border-bottom: 2px solid black
+      margin-bottom: 0.2em
+      border-top: thin solid black
 
     &--triple
-      border-bottom: 6px double black
+      margin-bottom: 0.2em
+      border-top: thin solid black
```
jlaasonen, line 50: These are all the same. There should only one style for a ruling line.

src/fragmentarium/ui/display/Transliteration.tsx
```diff
@@ -201,24 +201,38 @@ const lineComponents: ReadonlyMap<
...
+    for (let i = 1; i <= rulingsNumber; i++) {
+      items.push(
+        React.createElement('div', {
+          className: `Transliteration__RulingDollarLine--${line.number.toLowerCase()}`,
+          key: i
+        })
+      )
+    }
+    return items
+  }
```
jlaasonen, line 229: This seems a very convoluted way to create three elements.

- This could be a regular component, instead of returning a list of elements.
- There is no need to use `createElement`, you can just use `<div key=... className=...>`
- There is no need for an intermediate `items` variable. You can use `map` on a range of numbers.
- It does not make a sense to add `Transliteration__RulingDollarLine` element inside a `Transliteration__RulingDollarLine` element. The elements used for each line should have another name, e.g. `Transliteration__ruling`.

jlaasonen:
[Change thickness of rulings](https://trello.com/c/ufrL0M8T/402-change-thickness-of-rulings)

src/fragmentarium/ui/display/Display.sass
```diff
@@ -36,20 +36,12 @@
...
-    border-top: 2px solid black
-    height: 2px
-    margin: 0
-
-    &--double
-      border-bottom: 2px solid black
-
+    &--single,
+    &--double,
     &--triple
```
jlaasonen, line 42: Why three modifiers? The ruling style is always the same.

src/fragmentarium/ui/display/Transliteration.tsx
```diff
@@ -201,26 +201,37 @@ const lineComponents: ReadonlyMap<
...
     container,
-    { className: `Transliteration__${rulingDollarLine.type}` },
-    <hr
-      className={classNames([
-        element,
-        `${element}--${rulingDollarLine.number.toLowerCase()}`
-      ])}
-    />
+    { className: 'Transliteration__RulingDollarLine' },
+    [...Array(rulingsNumber).keys()].map((value: number) => {
```
jlaasonen, line 223: What's this? It is hard to see what it is supposed to do.

jlaasonen: @yCobanoglu please, do not resolve conversations without addressing the issue in some way

The line is very cryptic. You should think about ways to make it easier to understand.

jlaasonen: It seems that you are creating a range of numbers. Extracting a method with a descriptive name would make that more obvious, or maybe [someone already implemented such a method](https://lodash.com/docs/4.17.15#range).

PR #21, opened 2020-04-21
Use Auth0 SPA client

jlaasonen:


jlaasonen:
[User needs to click "Login" twice](https://trello.com/c/sBILSKaY/272-user-needs-to-click-login-twice)

PR #24, opened 2020-04-30
bibliography-search now also with container-title-short & collection-number

yCobanoglu:


src/bibliography/infrastructure/BibliographyRepository.test.ts
```diff
@@ -28,13 +28,13 @@ const testData: TestData[] = [
...
+    [[author, year, title]],
     apiClient.fetchJson,
     [entry],
     [
-      `/bibliography?author=${encodeURIComponent(
-        author
-      )}&title=${encodeURIComponent(title)}&year=${encodeURIComponent(year)}`,
+      `/bibliography?0=${encodeURIComponent(author)}&1=${encodeURIComponent(
+        year
+      )}&2=${encodeURIComponent(title)}`,
```
jlaasonen, line 37: Weren't you going to use an "array" parameter? This is same as "first", "second", third" but with different names.

yCobanoglu:
[Add Container-title-short to Bibliography Editor](https://trello.com/c/skV5pHJ7/318-add-container-title-short-to-bibliography-editor)

src/bibliography/application/BibliographyService.ts
```diff
@@ -25,14 +25,14 @@ export default class BibliographyService implements BibliographySearch {
   }
 
   search(query: string): Promise<readonly BibliographyEntry[]> {
-    const queryRegex = /^([^\d]+)(?: (\d{4})(?: (.*))?)?$/
+    const queryRegex = /^([^\d]+)(?: (\d{1,4})(?: (.*))?)?$/
```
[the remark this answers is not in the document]
jlaasonen: It does not work. Also, collection numbers are strings. It would be better to do the parsing in the API based on the which query is being executed.

src/bibliography/infrastructure/BibliographyRepository.test.ts
yCobanoglu: But this is how an array gets encoded automatically as url

src/bibliography/application/BibliographyService.ts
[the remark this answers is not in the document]
yCobanoglu: So all parsing in the backend ? That would mean doin the get request with a single paramter which is the search input string?

src/bibliography/infrastructure/BibliographyRepository.test.ts
yCobanoglu: Have you meant passig the array like this (as an extra parameter)
```
...
[
    `/bibliography`, [author, year, title], true,
]
...

src/bibliography/application/BibliographyService.ts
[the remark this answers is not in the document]
jlaasonen: Yes.

src/bibliography/infrastructure/BibliographyRepository.test.ts
jlaasonen: Like described here: https://falcon.readthedocs.io/en/stable/api/request_and_response.html#falcon.Request.params

, but if the parsing is done in the backend, it does not matter.

PR #26, opened 2020-05-18
Display bibliography notes

jlaasonen:


jlaasonen:
[Inline bibliography](https://trello.com/c/rnxNEwJT/274-inline-bibliography)

PR #28, opened 2020-05-28
referenceSearch in fragmentarium

yCobanoglu:


yCobanoglu:
There is still a pretty bad bug if you actually search something in the front page (like Fragment Number) and then once the results of the search are shown and you press search again you can see the bug on the userinterface and the url changes too. I dont know why seems like a React thing.

jlaasonen:
> There is still a pretty bad bug if you actually search something in the front page (like Fragment Number) and then once the results of the search are shown and you press search again you can see the bug on the userinterface and the url changes too. I dont know why seems like a React thing.

It is not a React thing. You are initializing the state of `SearchGroup` incorrectly and the state is not used to change fields. There are redundant states inside the search forms. You should add test for the `SearchGroup`

jlaasonen:
[Search Bibliography](https://trello.com/c/22hMomoO/319-search-bibliography)

src/fragmentarium/ui/search/NumberSearchForm.tsx
```diff
@@ -1,57 +1,33 @@
...
-      <Form onSubmit={this.submit}>
+      <Form>
         <Form.Group as={Row} controlId="number">
-          <Col sm={2} />
-          <Col sm={7}>
+          <Col sm={{ span: 10, offset: 2 }}>
             <Form.Control
               type="text"
-              value={this.state.number}
+              value={this.props.getState('number')}
```
jlaasonen, line 21: This will most likely mess up updating the component. Why not just pass the number as a prop?

src/fragmentarium/ui/search/TransliterationSearchForm.tsx
```diff
@@ -80,25 +56,20 @@ class TransliterationSearchForm extends Component<Props, State> {
           >
             <HelpTrigger overlay={TransliterationSearchHelp()} />
           </Col>
-          <Col sm={7}>
+          <Col sm={10}>
             <Form.Control
               as="textarea"
-              value={this.state.transliteration}
+              value={this.props.getState('transliteration')}
```
jlaasonen, line 62: Pass the transliteration as prop.

src/fragmentarium/ui/search/NumberSearchForm.tsx
```diff
@@ -1,57 +1,33 @@
...
-  }
-
-  submit = (event) => {
-    event.preventDefault()
-    this.props.history.push(
-      `/fragmentarium/search/?${stringify({
-        number: this.state.number,
-      })}`
-    )
+    this.props.onChange('number', event.target.value || '')
```
jlaasonen, line 11: You should only pass the value and not couple the implementation with something on the outside. Whoever defines the callback should make do the correct thing without the form knowing its internals.

src/fragmentarium/ui/search/ReferenceSearchForm.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+type Props = {
+  fragmentService
+  onChange(searchForm: string, searchQuery: string): void
+  getState(key: string): string
+}
+
+class ReferenceSearchForm extends Component<Props> {
+  onChangePage = (event) => {
+    this.props.onChange('pages', event.target.value || '')
```
jlaasonen, line 13: You should only pass the value and not couple the implementation with something on the outside. Whoever defines the callback should make do the correct thing without the form knowing its internals.

src/fragmentarium/ui/search/ReferenceSearchForm.tsx
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  render() {
+    return (
+      <Form>
+        <Form.Group as={Row} controlId="reference">
+          <Col sm={{ span: 5, offset: 2 }}>
+            <BibliographySelect
+              aria-label="BibliographyTitle"
+              value={this.props.getState('title')}
```
jlaasonen, line 23: Pass the value as a prop.

src/fragmentarium/ui/search/ReferenceSearchForm.tsx
```diff
@@ -0,0 +1,44 @@
...
+              searchBibliography={(query) =>
+                this.props.fragmentService.searchBibliography(query)
+              }
+            />
+          </Col>
+          <Col sm={5}>
+            <Form.Control
+              type="text"
+              aria-label="Pages"
+              value={this.props.getState('pages')}
```
jlaasonen, line 34: Pass the value as a prop.

src/fragmentarium/ui/search/TransliterationSearchForm.tsx
```diff
@@ -37,41 +35,19 @@ type State = {
...
-    )
-  }
-
-  componentDidUpdate(prevProps, prevState, snapshot) {
-    if (this.props.transliteration !== prevProps.transliteration) {
-      this.setState({
-        transliteration: this.props.transliteration || '',
-      })
-    }
+    this.props.onChange('transliteration', event.target.value || '')
```
jlaasonen, line 44: You should only pass the value and not couple the implementation with something on the outside. Whoever defines the callback should make do the correct thing without the form knowing its internals.

src/fragmentarium/ui/SearchGroup.tsx
```diff
@@ -1,28 +1,74 @@
...
+class SearchGroup extends Component<Props, State> {
+  constructor(props) {
+    super(props)
+    this.state = {
+      number: this.props.number || '',
+      title: this.props.title || '',
+      id: this.props.id || '',
+      pages: this.props.pages || '',
+      transliteration: this.props.transliteration || '',
+    }
```
jlaasonen, line 29: [Avoid copying props into state.](https://reactjs.org/docs/react-component.html#constructor)

jlaasonen:
> @jlaasonen SearchGroup test is missing and maybe there could be a way to reuse BibliographySelect (i currently duplicated it to use it in this PR). Other than that it is ready for CodeReview. Code Climate complains about BibliographySelect beeing a duplicate-

`BibliographySelect` should be reused. Why was it copy-pasted?

jlaasonen:
Won't fixed some of the duplicates. CodeClimate likes to complain a lot about frontend code.

src/fragmentarium/ui/SearchGroup.tsx
```diff
@@ -1,28 +1,74 @@
...
+        <ReferenceSearchForm
+          onChange={this.onChange}
+          getState={this.getState}
+          fragmentService={this.props.fragmentService}
+        />
+        <TransliterationSearchForm
+          onChange={this.onChange}
+          getState={this.getState}
+        />
+        <div className="SearchGroup__button-bar">
```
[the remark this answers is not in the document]
yCobanoglu: yes

src/fragmentarium/ui/FragmentButton.tsx
```diff
@@ -35,7 +35,7 @@ function FragmentButton({ query, history, children }: Props) {
 
   return (
     <>
-      <Button variant="primary" onClick={handleClick}>
+      <Button className="m-1" variant="primary" onClick={handleClick}>
```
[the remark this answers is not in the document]
yCobanoglu: adds margin to the button class

src/fragmentarium/ui/SearchGroup.tsx
```diff
@@ -1,28 +1,79 @@
...
+import { Button, ButtonToolbar, Col } from 'react-bootstrap'
+import { stringify } from 'query-string'
+import ReferenceSearchForm from 'fragmentarium/ui/search/ReferenceSearchForm'
+import {
+  FragmentariumSearchParams,
+  SearchGroupParams,
+} from 'fragmentarium/ui/fragmentariumSearch'
 
-import './SearchGroup.css'
+type State = SearchGroupParams
```
jlaasonen, line 15: Is there a need to create an alias for the type?

src/fragmentarium/ui/front-page/Fragmentarium.tsx
```diff
@@ -12,15 +12,15 @@ import './Fragmentarium.css'
 import { Session } from 'auth/Session'
 import { SectionCrumb } from 'common/Breadcrumbs'
 
-interface Props {
-  number: string | null | undefined
-  transliteration: string | null | undefined
-  fragmentService
-  fragmentSearchService
-}
+import { FragmentariumSearchParams } from '../fragmentariumSearch'
```
jlaasonen, line 15: Avoid relative paths.

src/fragmentarium/ui/front-page/Fragmentarium.tsx
```diff
@@ -12,15 +12,15 @@ import './Fragmentarium.css'
...
+
+type Props = FragmentariumSearchParams
```
jlaasonen, line 17: Is there a need to create an alias for the type?

src/fragmentarium/ui/fragmentariumSearch.ts
```diff
@@ -0,0 +1,12 @@
...
+  id: string | null | undefined
+  title: string | null | undefined
+  pages: string | null | undefined
+  transliteration: string | null | undefined
+}
+
+export interface FragmentariumSearchParams extends SearchGroupParams {
+  fragmentService
+  fragmentSearchService
+}
```
jlaasonen, line 12: Are these props or parameters? The naming suggests that they belong to specific components. If that is the case they should be defined along with the component. If it something more generic it should be named accordingly e.g. `FragmentSearchParameters`.

You could also consider having a full model for the parameters. Then e.g. the transliteration replace could be handled in the model instead of in the components.

src/fragmentarium/ui/search/FragmentariumSearch.tsx
```diff
@@ -9,15 +9,18 @@ import replaceTransliteration from 'fragmentarium/domain/replaceTransliteration'
 import './FragmentariumSearch.css'
 import { SectionCrumb, TextCrumb } from 'common/Breadcrumbs'
 import { Session } from 'auth/Session'
+import { FragmentariumSearchParams } from 'fragmentarium/ui/fragmentariumSearch'
+import ReferenceSearch from './ReferenceSearch'
+
+type Props = FragmentariumSearchParams
```
jlaasonen, line 15: Is there a need to create an alias for the type?

src/fragmentarium/ui/search/ReferenceSearchForm.tsx
yCobanoglu: I fixed this by defining multiple callbacks one for NumberSearchForm, ReferenceSearchForm and TransliterationForm.

src/fragmentarium/ui/SearchGroup.tsx
yCobanoglu: But we need some kind of State for storing the props so that we can manipulate them from the children classes (we cant manipulate the props directly)

jlaasonen: This was handled previously handled via the query string. You can leave the state for now if it does not cause any issues.

src/fragmentarium/ui/fragmentariumSearch.ts
yCobanoglu: i justs defined them in the corresponding components and removed this file

src/bibliography/ui/BibliographySelect.tsx
```diff
@@ -9,7 +9,7 @@ function createOption(entry) {
   return entry && entry.id
     ? {
         value: entry.id,
-        label: createLabel(entry),
+        label: entry.label ? entry.label : createLabel(entry),
```
jlaasonen, line 12: `BibliographyEntry` does not have `label`.

src/fragmentarium/ui/search/ReferenceSearchForm.tsx
```diff
@@ -0,0 +1,43 @@
+import React, { Component } from 'react'
+import { Col, Form, Row } from 'react-bootstrap'
+import BibliographySelect from '../../../bibliography/ui/BibliographySelect'
+
+type Props = {
+  fragmentService
+  onChangePages(value: any): void
+  onChangeBibliographyReference(value: any): void
+  valueBibReference: any
```
jlaasonen, line 9: Do not use `any`. There is a class for bibliography entries.

src/fragmentarium/ui/search/ReferenceSearchForm.tsx
```diff
@@ -0,0 +1,43 @@
+import React, { Component } from 'react'
+import { Col, Form, Row } from 'react-bootstrap'
+import BibliographySelect from '../../../bibliography/ui/BibliographySelect'
```
jlaasonen, line 3: Avoid relative paths.

src/fragmentarium/infrastructure/FragmentRepository.ts
```diff
@@ -90,6 +97,32 @@ class ApiFragmentRepository
     return this._fetch({ number })
   }
 
+  searchReference(id: string, pages: string): FragmentInfosPromise {
+    return this._fetch({ id, pages }).then((fragInfos) => {
+      const newFragInfo: any = produce(
```
jlaasonen, line 102: Do not use `any`. Is this supposed to be `FragmentInfo` or `FragmentInfo[]`? The whole method looks rather convoluted and could be refactored to be more readable.

PR #34, opened 2020-07-20
Validate user input

yCobanoglu:


jlaasonen:
Was it not possible to use `type="number"`?

yCobanoglu:
> Was it not possible to use `type="number"`?

It is possible. Then if a user would input letters nothing would happen ( no error message whatsoever). I would think the way it is now is clearer for the user.

yCobanoglu:
i have to implmement tests for searchgroup to pass coverage.

yCobanoglu:
[Search Bibliography](https://trello.com/c/22hMomoO/319-search-bibliography)

yCobanoglu:
@jlaasonen I copy/pasted some parts of BibliographySelect.test.tsx to test (BibliographySelect) it in SearchForms..test.tsx.
I also added an `aria-label: BibliographySelect` to adress it in the tests.

yCobanoglu:
> @jlaasonen I copy/pasted some parts of BibliographySelect.test.tsx to test (BibliographySelect) it in SearchForms.test.tsx.
> I also added an `aria-label: BibliographySelect` to use it in the tests.

I will try refactoring codeclimate complains

yCobanoglu:
@jlaasonen you have to mark the codeclimate issue as unfixable.

src/bibliography/ui/BibliographySelect.tsx
```diff
@@ -52,6 +52,7 @@ export default class BibliographySelect extends Component<
     return (
       <>
         <AsyncSelect
+          aria-label="BibliographySelect"
```
jlaasonen, line 55: There is already a label defined on the next line.

Labels should be meaningful to the users, not component names. [data-testid](https://testing-library.com/docs/dom-testing-library/api-queries#bytestid) should be used if an ID is needed for testing, but it is seldom needed, e.g. the BibliohraphySelect has a label which can be used.

src/fragmentarium/ui/SearchForms.test.tsx
```diff
@@ -0,0 +1,120 @@
...
+let transliteration
+let fragmentService
+let fragmentSearchService
+let session
+let container
+let element
+let history
+
+let entry
+let searchEntry
```
jlaasonen, line 31: Use types.

src/fragmentarium/ui/SearchForms.test.tsx
```diff
@@ -0,0 +1,120 @@
...
+          fragmentService={fragmentService}
+          fragmentSearchService={fragmentSearchService}
+        />
+      </SessionContext.Provider>
+    </Router>
+  )
+  container = element.container
+  await element.getByLabelText('Number')
+  await element.getByLabelText('Pages')
+  await element.getByLabelText('Transliteration')
```
jlaasonen, line 57: What's this for?

src/fragmentarium/ui/SearchForms.tsx
```diff
@@ -0,0 +1,230 @@
...
+  year: string | null | undefined
+  title: string | null | undefined
+  pages: string | null | undefined
+  transliteration: string | null | undefined
+  fragmentService
+  fragmentSearchService
+  history: History
+} & RouteComponentProps
+
+class SearchForms extends Component<Props, State> {
```
jlaasonen, line 32: Why the name is plural? There is only one form now.

src/fragmentarium/ui/SearchForms.tsx
```diff
@@ -0,0 +1,230 @@
...
+    }
+  }
+
+  onChange = (name: string) => (value): void => {
+    this.setState((prevState) => ({ ...prevState, [name]: value }))
+  }
+
+  onChangePages = (value: string): void => {
+    this.setState({ pages: value })
+    this.setState({ isValid: this.isValid(value) })
```
jlaasonen, line 55: Use only one `setState` call.

src/fragmentarium/ui/SearchForms.tsx
```diff
@@ -0,0 +1,230 @@
...
+    return !!(Number(pages) || !pages)
+  }
+
+  onChangeBibliographyReference = (event) => {
+    const newState = { ...this.state }
+    newState.referenceEntry.title = event.cslData.title || ''
+    newState.referenceEntry.id = event.cslData.id || ''
+    newState.referenceEntry.primaryAuthor = event.cslData.author[0].family || ''
+    newState.referenceEntry.year =
+      event.cslData.issued['date-parts'][0][0] || ''
```
jlaasonen, line 67: Do not mutate state. Use [Immer](https://immerjs.github.io/immer/docs/introduction).

src/test-support/test-bibliographySelect.ts
```diff
@@ -0,0 +1,19 @@
+import { changeValueByLabel, clickNth } from './utils'
+import BibliographyEntry from '../bibliography/domain/BibliographyEntry'
+import { RenderResult } from '@testing-library/react'
+
+export async function fillBibliographySelect(
+  entry: BibliographyEntry,
+  BibliographySelectLabel: string,
```
jlaasonen, line 7: Argument names should be camel-case.

src/fragmentarium/ui/SearchForms.tsx
```diff
@@ -0,0 +1,230 @@
...
+                  event: React.ChangeEvent<HTMLTextAreaElement>
+                ): void => this.onChange('transliteration')(event.target.value)}
+              />
+            </Col>
+          </Form.Group>
+        </Form>
+        <ButtonToolbar>
+          <Col sm={{ offset: 2 }}>
+            <Button
+              className="w-25 m-1"
```
jlaasonen, line 211: What's the purpose of `w-25`? It makes the button look weird.

It is also a bit odd that this button has `m-1`, but the two others do not.

src/bibliography/ui/BibliographySelect.tsx
yCobanoglu: Would it make sense to change the "aria-labelledby" to "aria-label" or why are we use "aria-labelledby" ?

jlaasonen: To tell which element is the label. If you change it `ReferenceForm` will get broken. If you absolutely have to go without a real label (can't remember if it was a requirement not have labels here) you can make it take either or.

In general, it is considered a bad practice not to have real labels. See e.g.:

- https://joshuawinn.com/ux-input-placeholders-are-not-labels/
- https://uxplanet.org/designing-more-efficient-forms-structure-inputs-labels-and-actions-e3a47007114f
- https://www.uxbooth.com/articles/the-new-rules-of-form-design/

src/fragmentarium/ui/SearchForms.test.tsx
yCobanoglu: `
...
let fragmentService: any
let fragmentSearchService: any
let session: any
let element: RenderResult
let history: any
...
`
have types any the rest has proper types

src/fragmentarium/ui/SearchForms.tsx
yCobanoglu: `w-25` adds padding to the button so the button is wider than the content (= text within the button, e.g. "search").  This makes all the button look more similar in width regardless of their content. I can change it (the search button will look smaller then the other two though.)

src/bibliography/ui/BibliographySelect.tsx
yCobanoglu: i decided to get rid of `aria-labelledyby' and use `aria-label` instead.
For `aria-labelled-by` there is a `label` (with text needed) but in `SearchForm.tsx` there is no need for a label with text. If the `label` is empty there is no way of accesing it in the test other than using the `labels id` which is bad practiced. That's why i got rid of `aria-labelled-by`. `aria-label` gets passed: `_.uniqueId('ReferenceForm-Document-')` which i access in the tests using regex `/ReferenceForm-Document-.*` via `changeValueByLabel`

src/fragmentarium/ui/SearchForms.tsx
yCobanoglu: I decided to leave it like that for now instead of changing it. It looks better in my opinion we can change it anytime afterwards

PR #36, opened 2020-08-25
Tei export

Hamster3001:


src/fragmentarium/ui/fragment/Download.tsx
```diff
@@ -10,12 +11,24 @@ export default function Download({
...
+    const teiUrl = URL.createObjectURL(
+      new Blob([TeiExport.teiExport(fragment)], {
+        type: 'text/plain',
+      })
+    )
+    setTei(teiUrl)
+    return (): void => {
+      URL.revokeObjectURL(teiUrl)
+    }
+
```
jlaasonen, line 25: This code is unreachable because of the `return` above.

src/fragmentarium/ui/fragment/Download.tsx
```diff
@@ -10,12 +11,24 @@ export default function Download({
...
   const [atf, setAtf] = useState<string>()
+  const [txml, setTei] = useState<string>()
   useEffect(() => {
+    const teiUrl = URL.createObjectURL(
+      new Blob([TeiExport.teiExport(fragment)], {
+        type: 'text/plain',
+      })
+    )
+    setTei(teiUrl)
+    return (): void => {
```
jlaasonen, line 22: You return here. The rest of the function is unreachable.

src/fragmentarium/ui/fragment/TeiExport.js
```diff
@@ -0,0 +1,117 @@
...
+      '  <xsl:template match="node()|@*">',
+      '    <xsl:copy><xsl:apply-templates select="node()|@*"/></xsl:copy>',
+      '  </xsl:template>',
+      '  <xsl:output indent="yes"/>',
+      '</xsl:stylesheet>',
+    ].join('\n'),
+    'application/xml'
+  )
+
+  var xsltProcessor = new XSLTProcessor()
```
jlaasonen, line 112: This breaks the tests and it is not recommended for production sites.

> This feature is non-standard and is not on a standards track. Do not use it on production sites facing the Web: it will not work for every user. There may also be large incompatibilities between implementations and the behavior may change in the future.

See: [XSLTProcessor](https://developer.mozilla.org/en-US/docs/Web/API/XSLTProcessor)

src/fragmentarium/ui/fragment/TeiExport.js
```diff
@@ -0,0 +1,117 @@
...
+  const end = '</sourceDesc></fileDesc></teiHeader><text><body>'
+
+  return res + start + title_stmt + pub_stmt + source_desc + end
+}
+
+function getEnd() {
+  return '</body></text></TEI>'
+}
+
+function getBody(fragment) {
```
jlaasonen, line 44: The nested for-if-if construct is rather hard to read and definitely too complex. If you extract methods for the emptines check and creating the p-element, you can simplify this function to one line e.g. `fragment.text.allLines.filter(isNotEmpty).map(getParagraph).join('')`.

jlaasonen:
[Export TEI-XML](https://trello.com/c/yjddm2Eh/415-export-tei-xml)

src/fragmentarium/ui/fragment/Download.tsx
```diff
@@ -37,29 +49,34 @@ export default function Download({
...
         href={json}
         download={`${fragment.number}.json`}
       >
-        Downlad as JSON File
+        Download as JSON File
       </Dropdown.Item>
-      <Dropdown.Item eventKey="5" disabled>
-        Downlad as TEI XML File
+      <Dropdown.Item
+        data-testid="download-xml"
```
jlaasonen, line 74: Would it be possible to replace the testid's with getRole? 
See: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-the-wrong-query

src/fragmentarium/ui/fragment/Download.tsx
```diff
@@ -2,6 +2,7 @@ import React, { useEffect, useState } from 'react'
 import { DropdownButton, ButtonGroup, Dropdown } from 'react-bootstrap'
 import _ from 'lodash'
 import { Fragment } from 'fragmentarium/domain/fragment'
+import * as TeiExport from './TeiExport.js'
```
jlaasonen, line 5: All code should be TypeScript. JavaScript is no longer allowed in master.

src/fragmentarium/ui/fragment/TeiExport.js
```diff
@@ -0,0 +1,138 @@
...
+  var res = ''
+  for (var j = 0; j < line.content.length; j++) {
+    var word = escapeXMLChars(line.content[j].value)
+    res += word
+    if (j !== line.content.length - 1) res += ' '
+  }
+  return res
+}
+
+function escapeXMLChars(word) {
```
jlaasonen, line 109: Do not capitalize acronyms, to be consistent with the rest of the codebase (including `prettifyXml`) below.

src/fragmentarium/ui/fragment/TeiExport.js
```diff
@@ -0,0 +1,138 @@
+export function teiExport(fragment) {
```
jlaasonen, line 1: Make a unit test for the XML conversion.

src/fragmentarium/ui/fragment/TeiExport.js
```diff
@@ -0,0 +1,138 @@
...
+  var res = ''
```
jlaasonen, line 2: Use descriptive and unshortened names. I.e `result`.

src/fragmentarium/ui/fragment/TeiExport.js
```diff
@@ -0,0 +1,138 @@
...
+  const header = getHeader(fragment)
+  const body = getBody(fragment)
+  const end = getEnd()
+
+  res += header
+  res += body
+  res += end
+
+  console.log(fragment)
```
jlaasonen, line 11: Remove debug logging.

src/fragmentarium/ui/fragment/TeiExport.js
```diff
@@ -0,0 +1,138 @@
...
+}
+
+function isNotEmpty() {
+  return (line) => line.type !== 'EmptyLine'
+}
+
+function getParagraph(line) {
+  var res = ''
+
+  if (line.type === 'TextLine') {
```
jlaasonen, line 84: You can use [a type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-differentiating-types) to the get correct type information inside the branches.
See: https://github.com/ElectronicBabylonianLiterature/ebl-frontend/blob/4d081ad499abea12f15c5dce18431e4055bb35b1/src/transliteration/domain/type-guards.ts

src/fragmentarium/ui/fragment/TeiExport.js
```diff
@@ -0,0 +1,138 @@
...
+    res += getLineContent(line)
+    res += '</p>'
+  }
+
+  return res
+}
+
+function getLineContent(line) {
+  var res = ''
+  for (var j = 0; j < line.content.length; j++) {
```
jlaasonen, line 101: Do not use generic one letter names (i, j, k, x, ...). I.e use `index` for indices.

src/fragmentarium/ui/fragment/TeiExport.js
```diff
@@ -0,0 +1,138 @@
...
+function getEnd() {
+  return '</body></text></TEI>'
+}
+
+function getBody(fragment) {
+  return fragment.text.allLines.filter(isNotEmpty()).map(getParagraph).join('')
+}
+
+function isNotEmpty() {
+  return (line) => line.type !== 'EmptyLine'
```
jlaasonen, line 78: There should be no need to return another function. See also type guard below.

jlaasonen (changes requested):
Remove commented-out code.

Add types. They will allow spotting bugs during compile, improve IDE suggestions, and make refactoring much easier.

src/fragmentarium/ui/fragment/Download.tsx
```diff
@@ -37,29 +49,34 @@ export default function Download({
...
+        Download as ATF
       </Dropdown.Item>
       <Dropdown.Item
         data-testid="download-json"
         eventKey="4"
         href={json}
         download={`${fragment.number}.json`}
       >
-        Downlad as JSON File
+        Download as JSON File
```
jlaasonen, line 71: 👍

jlaasonen:
There is a lot of string catenation going on. Have you tested it with a very large and complex transliteration?

Alternatives would include using `.map(...).join(...)` or [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals).

src/fragmentarium/ui/fragment/Download.tsx
Hamster3001: Changed, will be .tsx in next commit.

src/fragmentarium/ui/fragment/TeiExport.js
Hamster3001: changed.

src/fragmentarium/ui/fragment/TeiExport.js
Hamster3001: changed res to result

src/fragmentarium/ui/fragment/TeiExport.js
Hamster3001: removed for commit

src/fragmentarium/ui/fragment/TeiExport.js
Hamster3001: reject. makes code more unreadable and i as loop counter is widespread convention.

src/fragmentarium/ui/fragment/Download.tsx
Hamster3001: possible, but for what benefit? don't see the argument described in the text

src/fragmentarium/ui/fragment/TeiExport.js
Hamster3001: will do when all other issues are resolved (and the XML output is as desired)

src/fragmentarium/ui/fragment/TeiExport.js
Hamster3001: added notIsEmptyLine(line: AbstractLine) to type-guards and removed function

src/fragmentarium/ui/fragment/TeiExport.js
Hamster3001: using isTextLine(line) results in TS error: 
Property 'number' does not exist on type 'LineNumber | LineNumberRange'.
  Property 'number' does not exist on type 'LineNumberRange'.  TS2339

src/fragmentarium/ui/fragment/Download.tsx
Hamster3001: word in function getWord(word:any,escaped_word_value:string) does not appear to be of class word but of instance of object therefore using type "any"

jlaasonen: Using `any` should be avoided as it will hide errors and creates more warnings (Adding types is an ongoing process, so this is only a warning and not an error.). Contents of a line should be of type [Token](https://github.com/ElectronicBabylonianLiterature/ebl-frontend/blob/master/src/transliteration/domain/token.ts), not everything on a line is a `Word`. Token types are not classes, so they are plain `objects`, but should have the properties matching the interface.

src/fragmentarium/ui/fragment/TeiExport.js
jlaasonen: I do not get this error if a use `isTextLine` with the the latest version.

src/fragmentarium/ui/fragment/Download.tsx
jlaasonen: To give confidence that the component works when an actual user is using it. Test IDs are only available in the code and the might pass, even if the user is unable access the item. I.e. you are testing that the component has the correct test ID not that has the correct label/text/role.

Another reason to avoid test IDs is that they pollute the production code with test logic.

src/fragmentarium/ui/fragment/Download.tsx
jlaasonen: > Changed, will be .tsx in next commit.

The file should be `.ts`.  `.ts` is plain TypeScript. `.tsx` must be used only if the file contaings JSX.

src/fragmentarium/ui/fragment/TeiExport.js
jlaasonen: I disgagree that shortening automatically improves readability, but in a simple loop, this should not be an issue, and single letter names are used in [generics](https://www.typescriptlang.org/docs/handbook/generics.html). Although `i` should be used. If the function has multiple loops it needs refactoring. In the latest code `i` is also used in other functions, where better naming would be required or it must be kept strictly within the loop body (e.g. pass current and previous line).

The for-loop itself is not very readable. In addition to keeping track of the index and the result variable, the reader needs to figure out the purpose of the if. In the case above `map` + `join` would be more readable. It clearly communicates the intention of the code and gets rid of the intermediate variables. 

I do no net see many uses for the plain for-loop. [for..of](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of), `map`, `filter` and `forEach`, remove the boiler plate of accessing the element at index and can communicate the intent better. Arrays methods also have extra arguments for `index` and `array`.

src/fragmentarium/ui/fragment/TeiExport.tsx
```diff
@@ -0,0 +1,161 @@
+import { Fragment } from 'fragmentarium/domain/fragment'
+import { AbstractLine } from 'transliteration/domain/abstract-line'
+import { TextLine } from 'transliteration/domain/text-line'
```
jlaasonen, line 3: Remove unused imports.

src/fragmentarium/ui/fragment/TeiExport.tsx
```diff
@@ -0,0 +1,161 @@
...
+import { isTextLine, isEmptyLine } from 'transliteration/domain/type-guards'
+
+export function teiExport(fragment): string {
+  let final_output = ''
```
jlaasonen, line 7: Variable names should be camel case.

jlaasonen (changes requested):
Looks like eslint has gotten more lenient in some update. It used to raise an error on unused variables and wrong case.

src/fragmentarium/ui/fragment/Download.tsx
Hamster3001: Changed.

src/fragmentarium/ui/fragment/TeiExport.js
Hamster3001: Yes, that's why it is used in the current version comment above can be ignored.

src/fragmentarium/ui/fragment/TeiExport.js
Hamster3001: now using isEmtpyLine

src/fragmentarium/ui/fragment/TeiExport.js
Hamster3001: in the latest version, i is changed to previous_line. I disagree on using map, filter since it is 10 times slower (https://dev.to/henryjw/array-map-much-slower-than-for-loop-57if) than regular for loop and increases memory consumption due to copying arrays. We discussed performance lately and now you propose a construction that will significantly slow down the code, that does not appear reasonable to me.

src/fragmentarium/ui/fragment/TeiExport.tsx
Hamster3001: left in for the purpose of demonstration, will hopefully be used when type issue is solved

jlaasonen (commented):
It is not easy to see what some of the functions as supposed to do due to lots of for-loops, intermediate variables, and if-else blocks. It would be easy to mess things up when making changes. There should unit test(s) to make changing the code safer. I added some suggestions in comments, but the test should have the priority.

src/fragmentarium/ui/fragment/Download.tsx
Hamster3001: That's a better argument will change to role=button, will also change the other code in the written by sbd else

src/fragmentarium/ui/fragment/TeiExport.tsx
```diff
@@ -0,0 +1,161 @@
...
+    source_desc += '<p>Publication: ' + fragment.publication + '</p>'
+
+  const end = '</sourceDesc></fileDesc></teiHeader><text><body>'
+
+  return result + start + title_stmt + pub_stmt + source_desc + end
+}
+
+function getEnd(fragment: Fragment): string {
+  let notes = ''
+  if (fragment.notes) notes = '<note>' + fragment.notes + '</note>'
```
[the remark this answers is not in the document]
Hamster3001: no problem with minor refactor suggestions there is already an open issue regarding a unit test...

src/fragmentarium/ui/fragment/TeiExport.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+function getParagraph(fragment: Fragment): string {
+  let result = ''
+
+  const lines = fragment.text.allLines
+
+  for (let i = 0; i < lines.length; i++) {
+    const line = lines[i]
+
+    if (isEmptyLine(line)) continue
```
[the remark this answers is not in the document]
Hamster3001: That's not a claim but a fact. See link in other issue.

src/fragmentarium/ui/fragment/TeiExport.js
jlaasonen: Is the page too slow to use? Did you profile it? On different browsers?

The code should be made good first, only then fast. If using a more readable construction is not "too slow", there is need to do premature optimization. I raised the issue with string catenation because I was iterested in the effect it might have. It was investigated and no action deemed necessary, even if it seemed to use lot of memory compared to produced the XML size.

If it indeed is too slow then there might be other optimizations consider. See the comment about lodash.

src/fragmentarium/ui/fragment/TeiExport.tsx
Hamster3001: changed to finalOutput

src/fragmentarium/ui/fragment/TeiExport.tsx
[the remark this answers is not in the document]
Hamster3001: changed to: 

` const notes = (fragment.notes && fragment.notes.length > 0 && fragment.notes !== ' ') ? '<note>' + fragment.notes + '</note>' 
  : ''`

src/fragmentarium/ui/fragment/TeiExport.tsx
```diff
@@ -0,0 +1,161 @@
...
+    else if (isTextLine(line)) result += getTextLine(line, fragment)
+    else result += handleOtherLines(i, lines, line)
+  }
+  return result
+}
+
+function handleOtherLines(i, lines, line) {
+  let result = ''
```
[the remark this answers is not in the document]
Hamster3001: dont mind multiple returns changed

src/fragmentarium/ui/fragment/TeiExport.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+function getDollarLine(line) {
+  return '<note>' + line.prefix + getLineContent(line) + '</note>'
+}
+
+function getAtLine(i, lines, line) {
+  let result = ''
+  if (lines[i - 1] && lines[i - 1].prefix !== '@') result += '</lg><lg>'
+  result += '<note>' + getLineContent(line) + '</note>'
+  return result
```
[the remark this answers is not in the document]
Hamster3001: changed to: `let result = (previous_line && previous_line.prefix !== '@') ? '</lg><lg>' : ''
  result += '<note>' + getLineContent(line) + '</note>'`

src/fragmentarium/ui/fragment/TeiExport.tsx
```diff
@@ -0,0 +1,161 @@
...
+  )
+}
+
+function getLineContent(line): string {
+  let result = ''
+  for (let i = 0; i < line.content.length; i++) {
+    const word = line.content[i]
+    const escaped_word_value = escapeXmlChars(word.value)
+    if (line.type === 'TextLine') result += getWord(word, escaped_word_value)
+    else result += escaped_word_value
```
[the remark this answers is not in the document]
Hamster3001: changed to: `result += (isTextLine(line)) ? getWord(word, escaped_word_value) : escaped_word_value`

src/fragmentarium/ui/fragment/TeiExport.tsx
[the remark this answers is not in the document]
Hamster3001: @code issue changed to `
    if (isTextLine(line)) result += getTextLine(line, fragment)
    else if (!isEmptyLine(line)) result += handleOtherLines(lines[i - 1], lines, line)`

src/fragmentarium/ui/fragment/TeiExport.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+export function teiExport(fragment): string {
+  let final_output = ''
+  const header = getHeader(fragment)
+  const body = getBody(fragment)
+  const end = getEnd(fragment)
+
+  final_output += header
+  final_output += body
+  final_output += end
```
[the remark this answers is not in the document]
Hamster3001: change to `const finalOutput = header + body + end`

src/fragmentarium/ui/fragment/TeiExport.js
jlaasonen: I did quick tests with `performance.now()` on Edge (Other browsers might give different results.) and there appears to no performance difference. There were some outliers with long execution time, but the test fragment mostly took 1,6-2 s and the largest 7-9 s. 

10 times slower is only valid in the context of that micro-benchmark from over a year ago. Browsers might have been optimized after that, or the lost performance is won by reducing boilerplate code for managing the loops. These benchmarks can be useful to find possible optimizations after bottlenecks have been identified. Still, to avoid sub-optimization, the whole context must be taken into account when deciding if/what to optimize.

In this case, it appears that the code was made less maintainable for no benefit at all. Please abstain from making premature optimizations in the future.

However, the execution time of several seconds would call for optimizations. It appears that not prettifying could shave about 2 seconds from the largest fragment. Specific elements are always at the same level. Maybe the indentation could be added together with the element. Alternatively, the generation of the XML could be delayed until the user clicks the downloads button. This would remove the performance loss from the users who are not interested in the XML.

Hamster3001: Can you send/commit the version you tested against the current version somewhere?

jlaasonen: Got unit wrong in the last message 🤦 , it's hard to concentrate when the kid is at home. Should be milliseconds not seconds. Does not change the point though, expect that optimization should not be necessary.

Here's the commit https://github.com/ElectronicBabylonianLiterature/ebl-frontend/commit/66099b52a849c328da4e1f6b2527075cd8c4f320

Hamster3001: Ok, that appears to be more realistic ;) It does change the point of creating the export only on clicking the button though. Thanks for the commit i'll also do a performance test when i find the time.

src/fragmentarium/ui/fragment/Download.tsx
Hamster3001: change to getbyRole however the role needs to be "link" not button for the tests to work

jlaasonen (approved):
Add tests if possible. Should be merged by Friday.

PR #38, opened 2020-09-25
make search clearable after selecting 

yCobanoglu:
make async-select clearable. Clearing the async-select passes a `null` to the handleChange function.

### DIFF
--- a/src/bibliography/ui/BibliographySelect.tsx
+++ b/src/bibliography/ui/BibliographySelect.tsx
@@ -52,6 +52,7 @@ export default class BibliographySelect extends Component<
     return (
       <>
         <AsyncSelect
+          isClearable
           aria-label={this.props['aria-label']}
           placeholder="Name Year Title"
           cacheOptions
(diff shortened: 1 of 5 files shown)


yCobanoglu:
so in Fragmentarium search is clearable which is better. But in ReferenceEditor it is not because there we can just delete it instead of Clearing it.

src/bibliography/ui/BibliographySelect.tsx
```diff
@@ -45,13 +45,14 @@ export default class BibliographySelect extends Component<
     this.setState({
       selectedOption,
     })
-    this.props.onChange(selectedOption.entry)
+    this.props.onChange(selectedOption)
```
jlaasonen, line 48: Why do you leak implementation details to the event handler? The argument of `onChange` should be the new value (i.e. same type as the `value`), not any data structure specific to the current implementation.

jlaasonen (changes requested):

src/bibliography/ui/BibliographySelect.tsx
yCobanoglu: i created a new PR dealing with this. [https://github.com/ElectronicBabylonianLiterature/ebl-frontend/pull/41](url)

yCobanoglu:
New PR for this:
https://github.com/ElectronicBabylonianLiterature/ebl-frontend/pull/41

PR #48, opened 2020-10-26
Unfocus Header Labels when Clicking on ebl Logo in header BugFix

yCobanoglu:
If you click on the logo while your are in fragmentarium for example (fragmentarium label in the header is focused) the Header labels(fragmentarium) dont loose focus.
Now they do if you click on the logo.

### DIFF
--- a/src/Header.test.tsx
+++ b/src/Header.test.tsx
@@ -1,10 +1,10 @@
 import React from 'react'
 import { render, screen, act } from '@testing-library/react'
-import { MemoryRouter } from 'react-router-dom'
+import { MemoryRouter, Route, Router } from 'react-router-dom'
 import Header from './Header'
+import { createMemoryHistory } from 'history'
 import { AuthenticationContext, User } from 'auth/Auth'
 import userEvent from '@testing-library/user-event'
-
 let auth
 
 beforeEach(() => {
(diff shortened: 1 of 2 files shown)


jlaasonen:
What's this about? Please, use a descriptive title.  Do you mean *active* by *focus*?

src/Header.tsx
```diff
@@ -22,33 +22,58 @@ function EblLogo(): JSX.Element {
...
+function NavItem(props: {
+  href: string
+  title: string
+  eventKey: string
+}): JSX.Element {
   return (
     <Nav.Item>
       <LinkContainer to={props.href}>
-        <Nav.Link>{props.title}</Nav.Link>
+        <Nav.Link eventKey={props.eventKey}>{props.title}</Nav.Link>
```
jlaasonen, line 33: Is the `eventKey` required? [`activeKey` should also match the `href`.](https://react-bootstrap.github.io/components/navs/#nav-props)

src/Header.tsx
```diff
@@ -22,33 +22,58 @@ function EblLogo(): JSX.Element {
...
           </LinkContainer>
           <Navbar.Collapse id={id}>
-            <Nav className="mx-auto">
-              <NavItem href="/dictionary" title="Dictionary" />
-              <NavItem href="/corpus" title="Corpus" />
-              <NavItem href="/fragmentarium" title="Fragmentarium" />
-              <NavItem href="/bibliography" title="Bibliography" />
+            <Nav
+              activeKey={activeKey}
+              onSelect={(key) => setActiveKey(key!)}
```
jlaasonen, line 58: Fix eslint warning.

jlaasonen (changes requested):
Add tests.

src/Header.tsx
yCobanoglu: even better

src/Header.test.tsx
```diff
@@ -49,6 +50,16 @@ function commonTests(): void {
...
   })
 }
+describe('Unfocus Header Labels on clicking ebl Logo', () => {
+  beforeEach(async () => await renderHeader(true))
+
+  test('click Logo', () => {
+    userEvent.click(screen.getByText('Fragmentarium'))
+    userEvent.click(screen.getByTitle('electronic Babylonian Literature (eBL)'))
+    expect(screen.getByText('Fragmentarium')).not.toHaveClass('active')
+    expect(screen.getByText('Fragmentarium')).not.toHaveFocus()
```
jlaasonen, line 60: Is focus relevant here?

src/Header.tsx
```diff
@@ -33,18 +33,27 @@ function NavItem(props: { href: string; title: string }): JSX.Element {
 }
 
 export default function Header(): JSX.Element {
+  const [activeKey, setActiveKey] = useState<string | null>('/')
```
jlaasonen, line 36: There should be more comprehensive tests. You take the handling over from react-bootstrap and set the default active key to non-existing value. This behavior seems incorrect if the user navigates directly to somewhere else in than the root page. You should tests that correct item is still active in that case.

src/Header.test.tsx
yCobanoglu: nope

src/Header.tsx
yCobanoglu: Im not sure i understand. By navigating directly do you mean going directly to https://ebabylon.org/fragmentarium ? instead of https://ebabylon.org ? Should i test that ? It works in praxis.

yCobanoglu: I can write a test for making sure activeKey is set on "manual" redirect. As i said state wouldnt be set on direct navigation via link.

jlaasonen: > Im not sure i understand. By navigating directly do you mean going directly to https://ebabylon.org/fragmentarium ? instead of https://ebabylon.org ? Should i test that ? It works in praxis.

Yes, this used to work without manual state handling.

jlaasonen: > So it works as expected when using `null` as default key. When the user navigates directly the `activeKey` is null. When he afterwords decides to click on `Fragmentarium` for example the `activeKey` is `"/fragmnetarium"`. I would suspect thats how bootstrap deals with it. If we want to manually set `activeKey` on "direct navigation" (link for example). We would have to manually get the URL state with react-router and set it. It won't change anything funcitonality wise though (Everything is working currently).

Why we would have to set the active key manually if it would not change anything? Isn't the root cause that the `Nav` does not change active key on link click if a nav link has been clicked before? The current fix is a workaround for one specific link (`Nav` seems only concerned on the actual `Nav.Item`s not whatever may be inside `NavBar`). This could be sufficient now since there are not many cross section links. However, I was able to get two active by following links the glossary:
image

The real fix would be to make the `Nav` handle route changes correctly (It appears already to be sensitive to the route on the initial render), but that should be done within react-bootstrap. Do you want to open an issue?

yCobanoglu: im not open an issue but i added tests

jlaasonen (approved):

PR #67, opened 2021-01-14
Add suggestions to Corpus lemmatization

jlaasonen:


jlaasonen:
[Apply lemma suggestions](https://trello.com/c/XePL7Dv1/480-apply-lemma-suggestions)

PR #68, opened 2021-01-15
Word Export

Hamster3001:
Converts Display Mode HTML to a Word Document mainly using docx.js (https://www.npmjs.com/package/docx)


src/fragmentarium/ui/fragment/WordDownloadButton.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const handleClick = (event) => {
+    setIsLoading(true)
+    getWordBlob(fragment, wordService).then((blob) => {
+      saveAs(blob, `${fragment.number}.docx`)
+      setIsLoading(false)
+    })
+  }
+
+  return (
+    <>
```
jlaasonen, line 32: Fragment with one child is redundant.

src/fragmentarium/ui/fragment/WordDownloadButton.tsx
```diff
@@ -0,0 +1,51 @@
...
+  fragment: Fragment,
+  wordService: WordService
+): Promise<Blob> {
+  const wordDoc = await wordExport(fragment, wordService).then((doc) => {
+    return doc
+  })
+  const wordBlob: Blob = await Packer.toBlob(wordDoc).then((blob) => {
+    return blob
+  })
+  return wordBlob
```
jlaasonen, line 50: `then` with identity function is redundant. You should use Bluebird promises (see comment above). `await` will create a native JS promise which is not cancellable.

src/fragmentarium/ui/fragment/WordDownloadButton.tsx
```diff
@@ -0,0 +1,51 @@
...
+  children,
+}: Props): JSX.Element {
+  const [isLoading, setIsLoading] = useState(false)
+
+  const handleClick = (event) => {
+    setIsLoading(true)
+    getWordBlob(fragment, wordService).then((blob) => {
+      saveAs(blob, `${fragment.number}.docx`)
+      setIsLoading(false)
+    })
```
jlaasonen, line 28: This will cause warnings because the state could be changed after the component is unmounted. You should use [Bluebird](http://bluebirdjs.com/) promises and cancel when the component is unmounted. There is already `usePromiseEffect` hook tp handle that.

See: https://reactjs.org/blog/2015/12/16/ismounted-antipattern.html

src/fragmentarium/ui/fragment/WordExport.tsx
```diff
@@ -0,0 +1,428 @@
...
+function wrapWithMemoryRouter(component: JSX.Element): ReactElement {
+  return <MemoryRouter>{component}</MemoryRouter>
+}
+
+function getMainTableWithFootnotes(
+  table: JQuery,
+  footNotesLines: Paragraph[]
+): any {
+  table.hide()
+  $('body').append(table)
```
jlaasonen, line 86: jQuery should not run wild in a React application. React is unaware of the changes and can get confused. Can you make a `div` inside the component only make jQuery modifications there?

See: https://reactjs.org/docs/integrating-with-other-libraries.html

jlaasonen (changes requested):
Try to fix at least some of the CodeClimate issues. Especially `WordExport.tsx` is so large that GitHub won't show the diff by default.

src/fragmentarium/ui/fragment/Download.tsx
```diff
@@ -39,7 +46,7 @@ export default function Download({
       URL.revokeObjectURL(jsonUrl)
       URL.revokeObjectURL(teiUrl)
     }
-  }, [fragment])
+  }, [fragment, wordService])
```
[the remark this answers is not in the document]
Hamster3001: fixed, right, wasn't needed here

src/fragmentarium/ui/fragment/WordDownloadButton.tsx
Hamster3001: now has to children (jQuery ref div)

src/fragmentarium/ui/fragment/WordDownloadButton.tsx
Hamster3001: removed identity function, is there an example in the code where a Bluebird promise is used?

jlaasonen: E.g. `BibliographyEntryForm`, `cancellableFetch`.

Hamster3001: now wrapped in Bluebird Promises

src/fragmentarium/ui/fragment/WordExport.tsx
Hamster3001: now jquery only runs in "#jQueryContainer" which is passed as a jQuery-Object to WordExport

src/fragmentarium/ui/fragment/WordDownloadButton.tsx
Hamster3001: Done, as discussed in Slack.

jlaasonen (approved):

PR #69, opened 2021-01-15
reduce eslint warnings - 1

yCobanoglu:


src/fragmentarium/ui/SearchForm.test.tsx
```diff
@@ -53,34 +56,27 @@ async function renderSearchForms() {
...
-  session.isAllowedToTransliterateFragments.mockReturnValue(true)
+  fragmentService.searchBibliography = jest
+    .fn()
+    .mockReturnValue(Promise.resolve([searchEntry]))
+
+  fragmentSearchService.random = jest.fn()
+  fragmentSearchService.interesting = jest.fn()
+
+  session.isAllowedToReadFragments = jest.fn().mockReturnValue(true)
+  session.isAllowedToTransliterateFragments = jest.fn().mockReturnValue(true)
```
jlaasonen, line 67: Do not mutate services, especially when they are global. Class mocks already have all methods spied. Otherwise use `spyOn`.

src/fragmentarium/ui/SearchForm.test.tsx
```diff
@@ -5,9 +5,20 @@ import { factory } from 'factory-girl'
...
+import FragmentService from 'fragmentarium/application/FragmentService'
+import FragmentSearchService from 'fragmentarium/application/FragmentSearchService'
+import MemorySession from 'auth/Session'
+
+jest.mock('fragmentarium/application/FragmentService')
+
+const fragmentService = new (FragmentService as jest.Mock<FragmentService>)()
+const fragmentSearchService = new (FragmentSearchService as jest.Mock<
+  FragmentSearchService
+>)()
```
jlaasonen, line 20: Add a class mock for the service.

src/fragmentarium/ui/SearchForm.test.tsx
```diff
@@ -5,9 +5,20 @@ import { factory } from 'factory-girl'
...
+const session = new (MemorySession as jest.Mock<MemorySession>)()
```
jlaasonen, line 21: You could just create a real `MemorySession` as it does not require complex setup. Either pass the desired scopes or use spies.

If you want to use a class mock, remember to set it up.

jlaasonen, line 21: Is it safe to have global test doubles? Changes made by one test could leak to the next.

jlaasonen (changes requested):

src/fragmentarium/ui/SearchForm.test.tsx
jlaasonen: If `session` is an automatic mock, you should be able to just do `session.isAllowedToReadFragments..mockReturnValue(true)`. The mock constructor has to type `jest.Mock<jest.Mocked<MemorySession>>`.

If it is a normal class, you can spy it `jest.spyOn(session, 'isAllowedToReadFragments')` and then set the return value.
See:  https://jestjs.io/docs/en/jest-object#jestspyonobject-methodname

src/fragmentarium/ui/SearchForm.test.tsx
yCobanoglu: so spying like this:
```
const session = new (MemorySession as jest.Mock<MemorySession>)()
jest.spyOn(session, "isAllowed...").mockReturnValue 
```?

Trying:
```
const session = new (MemorySession as jest.Mock<<jest.Mocked<MemorySession>>)()
session.isAllowedToRead.mockReturnValue(true) 
```
errors with not a function.

But using similar pattern on fragmentService `( as jest.Mock<jested.Mock<FragmentService>>)` and then
using `fragmentService.statistics.mockReturnValue...` works fine...
Just MemorySession it errors but not on FragmentService.

Comment (yCobanoglu) [reply src/fragmentarium/ui/SearchForm.test.tsx:21]: Maybe i can just not Mock the MemorySession (just instantiate it with scope). Would this be ok ? Seems like it would make it more of an integration test then. It would be the easiset way, saving some lines of code.

Comment (yCobanoglu) [reply src/fragmentarium/ui/SearchForm.test.tsx:21]: i can put everything withing a **describe block**.

File: src/fragmentarium/ui/fragment/CuneiformFragment.test.tsx [hunk 4]
```diff
@@ -38,24 +43,23 @@ beforeEach(async () => {
...
+    Promise.resolve({ photoUrl: null })
+  )
+  fragmentSearchService = new (FragmentSearchService as jest.Mock<
+    FragmentSearchService
+  >)()
+  session = new (MemorySession as jest.Mock<jest.Mocked<MemorySession>>)()
+
+  session.isAllowedToTransliterateFragments.mockReturnValue(true)
+  session.isAllowedToLemmatizeFragments.mockReturnValue(false)
+  session.hasBetaAccess.mockReturnValue(true)
```
jlaasonen, line 62: Why is the value changed from `false` to `true`?

src/fragmentarium/ui/SearchForm.test.tsx
```diff
@@ -5,9 +5,23 @@ import { factory } from 'factory-girl'
...
+
+jest.mock('fragmentarium/application/FragmentService')
+jest.mock('auth/Session')
+
+const fragmentService = new (FragmentService as jest.Mock<
+  jest.Mocked<FragmentService>
+>)()
+const fragmentSearchService = new (FragmentSearchService as jest.Mock<
+  jest.Mocked<FragmentSearchService>
+>)()
```
jlaasonen, line 23: Automatic mock is not set up for `FragmentSearchService`.

jlaasonen (changes requested):

src/fragmentarium/ui/SearchForm.test.tsx
yCobanoglu: i forgot to import session as `jest.mock("auth/session") now it works. But i init Session as a real instance (not a mock) which i prefer.

jlaasonen (approved):
Fix the LGTM alert. Then you can merge.

PR #89, opened 2021-03-09
display script and create link for fragments in matches (lineToVec section)

yCobanoglu:


yCobanoglu:
[Modify match rulings](https://trello.com/c/rtZR8bbs/491-modify-match-rulings)

jlaasonen (approved):

PR #92, opened 2021-03-10
display notes of derived correclty

yCobanoglu:


yCobanoglu:
[Design expanded dictionary entry](https://trello.com/c/fNlwoqVt/31-design-expanded-dictionary-entry)

src/dictionary/ui/display/WordDisplayParts.tsx
```diff
@@ -22,54 +23,150 @@
...
 }
-interface Derivative {
+export function Logogram({
+  logogram,
+  notes,
+}: {
+  logogram: string[]
+  notes: string[]
+}): JSX.Element {
+  const MultipleLogograms = ({ logograms }): JSX.Element => (
```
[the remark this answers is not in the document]
jlaasonen: @yCobanoglu remove duplicate code.

src/dictionary/ui/display/WordDisplayParts.tsx
```diff
@@ -22,54 +23,150 @@ export function OtherForms({
...
+    <span>
+      {notes[0] && <Markdown text={notes[0]}>&nbsp;</Markdown>}
+      <Markdown text={logogram[0]}>&nbsp;</Markdown>
+      {logogram.length > 1 && (
+        <span>
+          (<MultipleLogograms logograms={logogram.slice(1)} />
+          )&nbsp;
+        </span>
+      )}
+      {notes[1] && (
```
jlaasonen, line 54: Why the other check is `length > 1` and the other `notes[1]`? Be consistent.

jlaasonen (changes requested):

src/dictionary/ui/display/WordDisplayParts.tsx
```diff
@@ -22,54 +23,163 @@ export function OtherForms({
...
+  separator: string
+  Component
+}
+export function Join({ list, separator, Component }: JoinProps): JSX.Element {
+  return (
+    <>
+      {' '}
+      {list.map((element, index) => (
+        <React.Fragment key={index}>
+          <Component {...element} />
```
jlaasonen, line 41: What's this? Why do you expand an element (which is actually `unknown` and not `Element`) to component props?

jlaasonen (changes requested):

src/dictionary/ui/display/WordDisplayParts.tsx
yCobanoglu: The thing is i am passing a Component and a list which has as elements the props to the Component (which was passed) so the props can be anything. This is a litte clunky because if the element of the list dont have the exact type as the props of the Component (to which they are passed) it doesn't work. That's why i had to make a similiar function **JoinMarkdown**. I don't really know a better way currenlty to implmement a join function for ReactComponents-

yCobanoglu: One way would be to get rid of the function and just use this everywhere  
```
<>
      {' '}
      {list.map((element, index) => (
        <React.Fragment key={index}>
          <Markdown text={element} />
          {index !== list.length - 1 ? separator : ''}
        </React.Fragment>
      ))}
    </>
```

jlaasonen: Please use descriptive names. I.e. do not call props element or list. 

Props cannot be anything. You can use generic types to enforce correct props.

Having a function is better than having duplicate joining everywhere. Would it work with elements instead of components? Could be possible to use `children` to pass the contents.

jlaasonen (approved):

PR #94, opened 2021-03-17
fix transliteration search

yCobanoglu:


yCobanoglu:
[Bug: Transform diacritics only works on the front page of the Fragmentarium](https://trello.com/c/boOfz3lH/489-bug-transform-diacritics-only-works-on-the-front-page-of-the-fragmentarium)

src/fragmentarium/ui/SearchForm.tsx
```diff
@@ -88,6 +89,9 @@ class SearchForm extends Component<Props, State> {
   }
   search = (event) => {
     event.preventDefault()
+    this.onChange('transliteration')(
+      replaceTransliteration(this.state.transliteration || '')
```
jlaasonen, line 93: Replace is also done in `FragmentariumSearch`. It should be only done once.

jlaasonen (changes requested):

src/fragmentarium/ui/SearchForm.tsx
yCobanoglu: Ok. I'm not sure how this should be done because SearchGroup in Fragmentarium Search isn't rerendered if the user clicks on search. If the user clicks search in Fragmentarium (start screen) the correct replaced Transliteration gets passed to SearchGroup but once the user is at FragmentariumSearch the SearchGroup doesn't get rerendered. I can either rerender it if for example the URL changes (similar to the withData which we could pass parameters which cause it to rerender) or stick to the solution which manipulates the value to the textarea input when the user clicks on Search.

jlaasonen: If you push the replaced transliteration to history. The replace in `FragmentariumSearch` would be redundant.

But rerendering the forms on URL changes is better for the user. Otherwise, the results and forms might not match if a search URL is pasted to the browser.

jlaasonen (approved):

PR #97, opened 2021-03-29
search form for signs

yCobanoglu:


yCobanoglu:
[Search for signs interface](https://trello.com/c/wDJwbcrk/514-search-for-signs-interface)

src/signs/ui/search/Signs.tsx
```diff
@@ -0,0 +1,54 @@
...
+import SignsSearchForm from 'signs/ui/search/SignsSearchForm'
+import SignsSearch from 'signs/ui/search/SignsSearch'
+import { SignQuery } from 'signs/domain/Sign'
+import _ from 'lodash'
+
+function parseQuery(value: string | null | string[]): string | undefined {
+  if (typeof value === 'string' && value !== '') {
+    return value
+  } else {
+    return undefined
```
jlaasonen, line 18: Do not return `undefined`, use `null`.

src/signs/ui/search/SignsSearch.tsx
```diff
@@ -0,0 +1,116 @@
...
+import { Col, OverlayTrigger, Popover, Row } from 'react-bootstrap'
+import compareAkkadianStrings from 'dictionary/domain/compareAkkadianStrings'
+import './Signs.css'
+
+interface Props {
+  signs: Sign[]
+  isIncludeHomophones: boolean
+}
+
+function signsSorted(signs: Sign[]): Sign[] {
```
jlaasonen, line 19: Function names should be verbs.

src/signs/ui/search/SignsSearchForm.tsx
```diff
@@ -0,0 +1,211 @@
...
+function SignsSearchHelp(): JSX.Element {
+  const Section = ({ label, text }) => (
+    <li>
+      <Row>
+        <Col xs={2} className="pr-1">
+          {label}
+        </Col>
+        <Col className="pl-0">{text}</Col>
+      </Row>
+    </li>
```
jlaasonen, line 24: Why put a row in a list item? It seems to make no sense.

jlaasonen (changes requested):
Does the included font have a compatible license?

If so, add acknowledgements to the readme. If not, remove it from the repository (e.g. with BFG).

src/signs/domain/Sign.ts
```diff
@@ -0,0 +1,120 @@
...
+export class Value {
+  readonly value: string
+  readonly subIndex: number | undefined
+
+  constructor(value: string, subIndex: number | undefined = undefined) {
+    this.value = value
+    this.subIndex = subIndex
+  }
+  get displaySubIndex(): string {
+    if (this.subIndex === undefined) {
```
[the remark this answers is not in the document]
yCobanoglu: Data from server is undefined if  server sends python none.

src/signs/ui/search/SignsSearchForm.tsx
yCobanoglu: It somehow forces the columns not to wrap.

src/signs/ui/search/Signs.tsx
yCobanoglu: So i can return null but i would have to convert it to undefined later on `sign.repository.search(stringify(signQuery))`. So the api decides on what to do depending on which query parameter are set and the rest has to be undefined.
Example currently api request: `signs?sign=ba&subIndex=1&value=ba`
This would break the api if using null because the api request would be : `signs?isComposite&isIncludeHomophones&sign=ba&subIndex=1&value=ba`

jlaasonen: You need to fix your request. Do not make code that depends on JS quirks. Hint: see [query-string](https://github.com/sindresorhus/query-string) documentation.

`undefined` is used when something is not defined. Given that defining something as `undefined` does not make sense and can make the code harder to understand. If you need to get rid of some properties you can filter with lodash.

src/dictionary/domain/compareAkkadianStrings.ts
```diff
@@ -57,3 +57,9 @@ export default function compareAkkadianStrings(
       .head() ?? Math.sign(replacedWord.length - anotherWordReplaced.length)
   )
 }
+export function cleanAkkadianString(akkadianString: string): string {
+  return akkadianString
+    .split('')
+    .map((signChar) => (alphabet.indexOf(signChar) > 0 ? signChar : ''))
```
jlaasonen, line 63: Why is the first alphabet cleaned out?

src/signs/domain/Sign.ts
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  get displayCuneiformSigns(): string {
+    return this.unicode.map((unicode) => String.fromCodePoint(unicode)).join('')
+  }
+  get displaySignName(): string {
+    return this.name.replace(/\|/g, '')
+  }
+
+  static fromJavascriptObject(signJavascriptObject: SignDto): Sign {
```
jlaasonen, line 109: Why not call it `fromDto`? That's what it does, right?

src/signs/ui/search/Signs.css
```diff
@@ -0,0 +1,18 @@
...
+  margin: 0 auto 2em auto;
+  max-width: 32em;
+}
+
+@font-face {
+  font-display: swap;
+  font-family: 'Assurbanipal';
+  font-style: normal;
+  font-weight: normal;
+  src: url('Assurbanipal.woff') format('woff'),
```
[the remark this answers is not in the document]
jlaasonen: Please fix this.

src/signs/ui/search/Signs.css
```diff
@@ -0,0 +1,18 @@
+.Signs-search {
+  margin: 0 auto 2em auto;
+  max-width: 32em;
+}
+
+@font-face {
```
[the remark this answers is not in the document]
jlaasonen: Please fix this.

src/signs/ui/search/Signs.css
```diff
@@ -0,0 +1,41 @@
...
+  font-display: swap;
```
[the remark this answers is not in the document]
jlaasonen: Please fix this.

jlaasonen (changes requested):

src/signs/ui/search/SignsSearchForm.tsx
jlaasonen: Lists should have nothing to do with columns or wrapping. Use a proper way of making it not-wrapping instead of relying on hacks. Not sure exactly what you are seeking to achieve but a table would provide fixed columns, or you could just set the width for the label.

src/signs/ui/search/Signs.css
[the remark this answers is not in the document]
yCobanoglu: this seems to be a codeclimate bug. Font display does exist on `@font-face`  [font-display](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face)

src/dictionary/domain/compareAkkadianStrings.ts
yCobanoglu: comparing them errors on non-akkadian letters and that 's why i make sure there are only akkadian letters (for comparing, sorting later on).

jlaasonen (approved):

PR #99, opened 2021-04-15
fix sup/sub markdown bug

yCobanoglu:


yCobanoglu:
[Error: Renderer for type `sup` not defined or is not renderable](https://trello.com/c/x0s0bkTo/516-error-renderer-for-type-sup-not-defined-or-is-not-renderable)

src/common/InlineMarkdown.tsx
```diff
@@ -13,6 +13,11 @@ export default function InlineMarkdown({
       plugins={[remarkSubSuper]}
       disallowedTypes={['paragraph']}
       unwrapDisallowed
+      renderers={{
+        paragraph: 'span',
```
jlaasonen, line 17: This seems redundant due `disallowedTypes` and `unwrapDisallowed`.

jlaasonen: @yCobanoglu how was this resolved? There are no changes and no comment explaining why it has to be so.

jlaasonen (changes requested):
See comment.

src/common/InlineMarkdown.tsx
yCobanoglu: ok now i removed paragraph span

jlaasonen (approved):

PR #102, opened 2021-04-26
Change lemmatizer and alignment dropdowns to modals

jlaasonen:
Overlays are too glitchy.


jlaasonen:
[Issues with Lemmatizer and aligner](https://trello.com/c/doC7cMDi/520-issues-with-lemmatizer-and-aligner)

PR #103, opened 2021-04-30
New url in export

jlaasonen:


jlaasonen:
[Change URL in exports and elsewhere](https://trello.com/c/wsNoatdV/524-change-url-in-exports-and-elsewhere)

PR #108, opened 2021-05-17
Signs

yCobanoglu:


yCobanoglu:
@jlaasonen 
This is how the markdown is converted to HTML. Once its HTML i manually parse the sub- and subscripts. (The remark-plugin **remark-sub-super** which we used earlier doesn't work. In their github they say you can just add it with` .use(remarkSubSuper)`. But it didn't work.

These library are needed for the conversion:
rehype-raw,
rehype-stringify,
remark-html,
remark-parse,
remark-rehype,
unified
```
unified()
      .use(remarkParse)
      .use(remark2rehype, { allowDangerousHtml: true })
      .use(raw)
      .use(stringify)
      .process(mesZlFormatted, function (err, file) {
        if (err) throw err
        setMesZlBody(subSup(String(file)))
      })
      
const subSup = (mesZL) =>
    mesZL
      .replace(/\^([^\^]*)\^/g, '<sup>$1</sup>')
      .replace(/~([^~]*)~/g, '<sub>$1</sub>')      
```

Once transformed the result (HTML string) is displayed via **dangerouslySetInnerHTML**.
Alternatively we could use [remark-react](https://github.com/remarkjs/remark-react) to make it "secure".

src/signs/ui/search/MesZLContent.tsx
```diff
@@ -0,0 +1,93 @@
...
+}: {
+  mesZl: string
+}): JSX.Element | null {
+  const [mesZlHead, setMesZlHead] = useState<string | null>(null)
+  const [mesZlBody, setMesZlBody] = useState<string | null>(null)
+
+  const mesZlLines = mesZl.split('\n')
+
+  const mesZlFormatted = mesZlLines
+    .slice(1, 6)
```
yCobanoglu, line 21: @jlaasonen Enrique decided just to show 6 Lines for now until the expanded Sign Display is finished. Otherwise it would be too many lines for the popover. (We cut at Line 6 and display a ... for now).

yCobanoglu:
sign_duplicate [PR](https://github.com/ElectronicBabylonianLiterature/ebl-frontend/pull/114) was used in place of this PR

PR #112, opened 2021-05-28
Do not suggest alignments if an alignment exits.

jlaasonen:


jlaasonen:
[Modification to alignment](https://trello.com/c/0a4i5Qln/527-modification-to-alignment)

PR #119, opened 2021-06-10
Expanded sign view

yCobanoglu:


src/common/markdownToHtml.tsx
```diff
@@ -0,0 +1,67 @@
...
+  ...props
+}: Props): JSX.Element | null {
+  const [html, setHtml] = useState<string>('')
+  const [isReady, setIsReady] = useState(false)
+  useEffect(() => {
+    ;(async () => {
+      setHtml(await convertMarkdownToSanitizedHtml(markdown))
+      setIsReady(true)
+    })()
+  }, [html])
```
jlaasonen, line 56: > React Hook useEffect has a missing dependency: 'markdown'. Either include it or remove the dependency array. If 'setHtml' needs the current value of 'markdown', you can also switch to useReducer instead of useState and read 'markdown' in the reducer

The effect sets `html`, there should be no need to depend on it.

You could use `withData` instead of custom effect.

src/signs/ui/display/LogogramWord.tsx
```diff
@@ -0,0 +1,36 @@
...
+        <Link to={`/dictionary/${word._id}`}>
+          {attested}
+          {lemma}
+        </Link>
+      ) : (
+        <em>{`${attested}${lemma}`}</em>
+      )}
+      <span>
+        {word.homonym && ` ${word.homonym}`}
+        {word.guideWord && `, “${word.guideWord}”`}
```
jlaasonen, line 22: Words should always have guide word and homonym.

src/signs/ui/display/SignHeading.tsx
```diff
@@ -0,0 +1,87 @@
...
+          <span className={'text-secondary'}>{description}</span>
+        </Col>
+      </Row>
+    </Col>
+  )
+
+  return (
+    <Row>
+      <Col xs={3}>
+        <h3>{signName}</h3>
```
jlaasonen, line 36: Why does `h3` appear before `h2`?

src/signs/ui/display/SignDisplay.tsx
```diff
@@ -1,137 +1,42 @@
...
-        </Col>
-      </Row>
-    </>
-  )
-}
-
-function SignDisplay({ sign }: { sign: Sign }): JSX.Element {
   return (
     <Container>
       <SignHeading
```
jlaasonen, line 22: Add breadcrumbs.

src/signs/ui/search/MesZLContent.tsx
```diff
@@ -55,75 +26,53 @@ export default function MesZlContent({
...
-          </ExternalLink>
         </div>
-      </>
-    )
-  } else {
-    return null
-  }
+      )}
+      <div className="text-center border border-dark">
+        <strong>From</strong>
```
jlaasonen, line 56: Should this be a heading and not `strong`?

src/common/markdownToHtml.tsx
```diff
@@ -0,0 +1,67 @@
...
+  //unified() outputs the HTML within a <p>...</p>, we remove it
+  return DOMPurify.sanitize(html.slice(3, html.length - 4))
+}
+
+type Container = 'div' | 'span'
+interface Props extends PropsWithChildren<any> {
+  container?: Container
+  markdown: string
+}
+export function ContainerWithInnerHtml({
```
jlaasonen, line 44: The name is confusing. The name implies implementation detail, but the component is actually used to render Markdown.

What's the difference between the `Markdown` component? Is it necessary to have three components for rendering Markdown? The new components could replace the existing uses of `ReactMarkdown`.

jlaasonen (changes requested):

src/common/markdownToHtml.tsx
yCobanoglu: but this is taking care of a mix of html (inline css) and markdown ReactMarkdown deals with regular markdown.

src/signs/ui/search/MesZLContent.tsx
yCobanoglu: strong seems correct i checked with the pdf from enrique.

src/common/markdownToHtml.tsx
yCobanoglu: i renamed it

yCobanoglu: we have InlineMarkdown and Markdown which are both wrappers around ReactMarkdown adding some custom features. And we now have MarkdownAndHtml which takes Markdown + Html + Inline CSS styles and returns a Html string.

src/signs/ui/search/MesZLContent.tsx
jlaasonen: Why is strong correct? The PDF has nothing to do with semantic HTML. Strong, em etc. should not be used for styling. There is CSS for that.

src/common/markdownToHtml.tsx
jlaasonen: Do we need all three? Looks like your implementation can also inline (i.e. `container="span"`). Could be confusing for the users to have two flavours of markdown. E.g. https://www.ebl.lmu.de/corpus/L/99/99 introduction contains `^Um^` which is not formatted.

jlaasonen (approved):
It's good enough for now.

src/signs/ui/search/MesZLContent.tsx
yCobanoglu: ok

src/common/markdownToHtml.tsx
yCobanoglu: i can merge InlineMarkdown and Markdown in a new PR. But MarkdownAndHtmlToHtml is needed for markdown + html + inline css to html.

PR #124, opened 2021-06-16
Expanded sign meszl section

yCobanoglu:
deploy this first and then add the Sign Information and the rest.


src/common/HelpTrigger.tsx
```diff
@@ -1,12 +1,14 @@
...
-): JSX.Element {
+type Props = {
+  className?: string
+} & Omit<OverlayTriggerProps, 'children'>
+
+function HelpTrigger({ className = '', ...props }: Props): JSX.Element {
   return (
     <OverlayTrigger placement="right" {...props}>
-      <i className="fas fa-info-circle" />
+      <i className={`${className ?? className} fas fa-info-circle`} />
```
jlaasonen, line 11: `?? className` is redundant. Did you mean to use something as default?

You can use [classnames](https://jedwatson.github.io/classnames/) to build the string.

jlaasonen (changes requested):
Add acknowledgements for the fonts to the readme,  or if they are not compatible with our license remove them and put to DB.

src/signs/ui/display/signDisplay.css
```diff
@@ -0,0 +1,37 @@
...
+.old-babylonian-cursive {
+  font-family: 'Old-Babylonian Cursive', Junicode, serif;
+}
+
+.hittite {
+  font-family: 'Hittite', Junicode, serif;
+}
+.neo-babylonian {
+  font-family: 'Neo-Babylonian', Junicode, serif;
+}
```
jlaasonen, line 13: What about using the BEM conventions?

src/signs/ui/display/SignDisplay.tsx
```diff
@@ -1,10 +1,121 @@
...
+
+function SignHeading({
+  signName,
+  cuneiformLetters,
+}: {
+  signName: string
+  cuneiformLetters: string
+}): JSX.Element {
+  const CuneiformFonts = ({
+    font = undefined,
```
jlaasonen, line 33: Use `null` instead of `undefined`.

The font is optional why set the default to a null value?

src/signs/ui/search/MesZLContent.tsx
```diff
@@ -85,7 +87,7 @@ export default function MesZlContent({
         <div
           dangerouslySetInnerHTML={{ __html: DOMPurify.sanitize(mesZlBody) }}
         />
-        {mesZl.split('\n').length > 7 && (
+        {mesZl.split('\n').length > 7 && cutOff !== -1 && (
```
jlaasonen, line 90: What's special about 7 and -1? Shouldn't the check be if the length is larger than cut off?

src/common/HelpTrigger.tsx
yCobanoglu: Not as default but to have the option to pass classNames to the component as props..

src/signs/ui/display/signDisplay.css
yCobanoglu: ok

src/signs/ui/display/SignDisplay.tsx
yCobanoglu: i made the default to an empty string.

src/signs/ui/search/MesZLContent.tsx
yCobanoglu: It should check if the length is larger than the cut off. But passing -1 will not cut it off at all.

src/signs/ui/display/SignDisplay.tsx
```diff
@@ -1,10 +1,124 @@
...
+  mesZl,
+}: {
+  signName: string
+  mesZl: string
+}): JSX.Element | null {
+  return (
+    <>
+      <Row>
+        <Col>
+          <h1>&#8544;. MesZL</h1>
```
jlaasonen, line 93: Headings in the page content should start with `h3`. `h1` is the eBL heading and `h2` is the page heading.

src/signs/ui/search/MesZLContent.tsx
```diff
@@ -85,7 +87,7 @@ export default function MesZlContent({
         <div
           dangerouslySetInnerHTML={{ __html: DOMPurify.sanitize(mesZlBody) }}
         />
-        {mesZl.split('\n').length > 7 && (
+        {mesZl.split('\n').length > cutOff && cutOff !== -1 && (
```
jlaasonen, line 90: Avoid using magic numbers and using special values. The code is easier to understand if the domain is `number | null` vs `number` with special behaviour for some number(s), but wouldn't it be simplest just to set the default to `Infinity`?

What would happen if `cutOff` is `-2`?

jlaasonen (changes requested):

src/signs/ui/display/SignDisplay.tsx
```diff
@@ -1,10 +1,124 @@
...
+          <span className={'text-secondary'}>{description}</span>
+        </Col>
+      </Row>
+    </Col>
+  )
+
+  return (
+    <Row>
+      <Col xs={3}>
+        <h1>{signName}</h1>
```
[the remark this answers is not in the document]
jlaasonen: Please fix.

jlaasonen (changes requested):

jlaasonen (approved):

PR #132, opened 2021-07-02
Coding Conventions

yCobanoglu:


yCobanoglu:
@jlaasonen also we could rename the fixtures to factories. `signFactory.ts` instead `sign-fixture.ts` for example.

README.md
```diff
@@ -79,6 +79,11 @@ yarn lighthouse <url>
 
 [bluebird](http://bluebirdjs.com) promises are used whenever a cancellable promise is needed. E.g. when loading data to components (see [isMounted is an Antipattern](https://reactjs.org/blog/2015/12/16/ismounted-antipattern.html)). bluebird is compatible with native JavaScript promises, but care should taken that a bluebird promise is always used when `Promise.cancel()` is needed.
 
+## CSS Convention
+
+File names have lower camel case and the same name as the TSX File they are used in.
```
jlaasonen, line 84: Why change the case instead of matching the TSX?

yCobanoglu: ok that would work too. So camelcase same name as the tsx.
Also: `SignDisplay.css` then the bem should be `sign-heading__description` instead of `sign-display__sign-heading__description` we could leave out the `sign-display` as it will be the same as the file name ?

jlaasonen: Yes `sign-display__sign-heading__description` has two elements which is not valid. The question is "is sign-heading a valid block?".

jlaasonen:
@yCobanoglu I copied the existing conventions and added conventions for all file names. If you do not have further comments you can merge this.

PR #175, opened 2021-11-03
y instead ctrl

yCobanoglu:


src/fragmentarium/ui/image-annotation/annotation-tool/FragmentAnnotation.tsx
```diff
@@ -102,11 +106,27 @@ function FragmentAnnotation({
...
+          setIsChangeExistingModeButtonPressed(true)
+        }
+      }
+    },
+    [setIsChangeExistingModeButtonPressed]
+  )
+
+  const onReleaseButton = useCallback(
+    (event) => {
+      if (event.keyCode === 89) {
```
jlaasonen, line 124: Extract constants for magic numbers.

src/fragmentarium/ui/image-annotation/annotation-tool/FragmentAnnotation.tsx
```diff
@@ -102,11 +106,27 @@ function FragmentAnnotation({
...
-    if (event.keyCode === 27) {
-      reset()
-    }
-  }, [])
+  const onPressingDown = useCallback(
+    (event) => {
+      if (event.keyCode === 27) {
+        reset()
+      } else {
+        if (event.keyCode === 89) {
```
jlaasonen, line 114: You can use else-if or switch instead of nesting

src/fragmentarium/ui/image-annotation/annotation-tool/Help.tsx
```diff
@@ -11,8 +11,7 @@ export default function Help(): JSX.Element {
             <b>Esc</b> Reset
           </li>
           <li>
-            <b>Ctrl + Select Annotation:</b> Change Reading on existing
-            Annotation
+            <b>y + Select Annotation:</b> Change Reading on existing Annotation
```
jlaasonen, line 14: Why `y`, feels less natural than control?

jlaasonen (changes requested):

src/fragmentarium/ui/image-annotation/annotation-tool/Help.tsx
yCobanoglu: can we leave it like that for now right click is problematic because it looks like i would have to "catch" the right click and return the coordinates in relative fashion to the internal annotation handler for click.

jlaasonen (approved):

PR #179, opened 2021-11-11
fix disable annotations on pressing shift

yCobanoglu:


src/signs/infrastructure/SignRepository.ts
```diff
@@ -32,7 +32,7 @@ class SignRepository {
     if (token.isSignPossiblyExisting()) {
       return this.search({
         value: token.name.toLowerCase(),
-        subIndex: token.subIndex as number,
+        subIndex: token.subIndex || 1,
```
jlaasonen, line 35: This is still incorrect. Remember the discussion about sub-indices earlier.

jlaasonen (approved):

PR #223, opened 2022-03-14
Fix margins agi

ejimsan:


src/corpus/ui/Introduction.tsx
```diff
@@ -47,7 +47,15 @@ export default function Introduction({
...
-      <ReactMarkdown className="text-view__markdown" source={intro} />
+      {intro.split('\n').map((paragraph, index) => (
+        <ReactMarkdown
+          className="text-view__markdown"
+          source={paragraph}
+          key={index}
+        >
+          {paragraph}
+        </ReactMarkdown>
+      ))}
```
jlaasonen, line 58: This has nothing to do with agi margins. Please remove.

Also, markdown already has support for paragraphs.

jlaasonen (changes requested):

PR #229, opened 2022-03-29
Display word info

jlaasonen:


jlaasonen:
[Display word info](https://trello.com/c/nLS4IqDn/545-display-word-info)

PR #241, opened 2022-04-22
Make testDelegation easier to understand

jlaasonen:


jlaasonen:
It should be possible to add even more typing to `TestData`, e.g.:

```typescript
export class TestData<
  S,
  M extends keyof S = any,
  T = any,
  Y extends any[] = any[]
> {
  constructor(
    public method: M,
    public params: S[M] extends (...args: any[]) => any
      ? Parameters<S[M]>
      : never,
    public target: jest.Mock<T, Y> | jest.MockInstance<T, Y>,
    public expectedResult: unknown,
    public expectedParams?: Y,
    public targetResult?: T
  ) {}
}
```

But some tests won't compile anymore due to incorrect types.

jlaasonen:
A fluent builder API would also be possible.

```typescript
when('search')
  .isCalledWith(['Alba Cecilia 1998 The Qualifications'])
  .thenCall(bibliographyRepository.search)
  .andReturn([resultStub])
```

PR #242, opened 2022-04-25
fix warnings

yCobanoglu:


src/fragmentarium/ui/images/CdliImages.test.tsx
```diff
@@ -21,8 +26,9 @@ beforeEach(async () => {
 })
 
 describe('CdliImages', () => {
-  beforeEach(() => {
+  beforeEach(async () => {
     renderImages()
+    await waitForElementToBeRemoved(() => screen.getByLabelText('Spinner'))
```
jlaasonen, line 31: Spinners are used a lot. There could be `waitForSpinners` function.

src/common/ErrorBoundary.test.tsx
```diff
@@ -31,19 +34,24 @@ describe('Children throw an error', () => {
...
   })
 
   it('Clicking report button opens report dialog', async () => {
     clickNth(element, 'Send a report', 0)
 
     expect(errorReportingService.showReportDialog).toHaveBeenCalled()
   })
+  afterEach(async () => {
+    spy.mockRestore()
+  })
```
jlaasonen, line 54: Why do you need to restore manually and multiple times?

yarn.txt
```diff
@@ -0,0 +1,3 @@
+yarn run v1.22.17
+$ react-scripts test --watchAll=false
+info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```
jlaasonen, line 3: What's this for?

src/common/ErrorBoundary.test.tsx
```diff
@@ -10,8 +10,11 @@ describe('Children throw an error', () => {
   let element
   let error
   let errorReportingService
+  let spy
 
   beforeEach(async () => {
+    spy = jest.spyOn(console, 'error')
+    spy.mockImplementation()
```
jlaasonen, line 17: Now that `error` is spied, you could also verify it was called or not called correctly.

src/fragmentarium/ui/images/CdliImages.test.tsx
yCobanoglu: @jlaasonen i would have to pass in screen and the function wouldn't reduce the code much can i just leave it for now ?

jlaasonen: It is not about reducing code. The same logic is duplicated in multiple places. If the spinner is changed all of them have to be updated. If there is a function to wait for the spinner only it has to be changed.

jlaasonen (approved):

src/fragmentarium/ui/images/CdliImages.test.tsx
yCobanoglu: alright then i will do it

PR #255, opened 2022-05-11
Sundry format changes

ejimsan:


fsimonjetz (approved):
Looks good to me.

src/bibliography/ui/ReferenceList.tsx
```diff
@@ -28,13 +28,11 @@ export default function ReferenceList({
...
-      {_.isEmpty(references) ? (
-        <p>No references</p>
-      ) : (
-        groupReferences(references).map(([type, group]) => (
-          <ReferenceGroup key={type} references={group} />
-        ))
-      )}
+      {_.isEmpty(references)
+        ? null
+        : groupReferences(references).map(([type, group]) => (
```
yCobanoglu, line 33: In the Fragment Editor "No References" won't be shown because the component is used in two places (Editor and SearchResults)

yCobanoglu, line 33: In the Fragment Editor "No References" won't be shown because the component is used in two places (Editor and SearchResults)

src/corpus/ui/ManuscriptPopover.tsx
```diff
@@ -60,7 +60,9 @@ export default function ManuscriptPopOver({
           {manuscript.period.name} {manuscript.period.description}
           <br />
         </p>
-        <ManuscriptReferences references={manuscript.references} />
+        <span className="manuscript-popover__references">
+          <ManuscriptReferences references={manuscript.references} />
+        </span>
```
yCobanoglu, line 65: span should go inside ManuscriptReferences (unless its used in multiple places and only this one needs the span this ceratin way)

yCobanoglu, line 65: span should go inside ManuscriptReferences (unless its used in multiple places and only this one needs the span this ceratin way)

PR #256, opened 2022-05-16
display fragmentarium matching Lines in display mode

yCobanoglu:


ejimsan:
[Use display mode in Search](https://trello.com/c/nkFK6XRn/405-use-display-mode-in-search)

PR #264, opened 2022-06-14
Uncertain fragments check optimization

khoidt:


khoidt:
[Create new endpoint for uncertain fragments in Chapters](https://trello.com/c/Hg3Lx7BV/598-create-new-endpoint-for-uncertain-fragments-in-chapters)

src/corpus/ui/Chapters.tsx
```diff
@@ -44,12 +44,19 @@ const Manuscripts = withData<
   {
     uncertainFragments: readonly UncertainFragment[]
     textService
```
fsimonjetz, line 46: Maybe add types `TextService` and `FragmentService`?

src/corpus/ui/Chapters.tsx
```diff
@@ -169,7 +176,17 @@ const Manuscripts = withData<
                   <ul className="list-of-manuscripts__uncertain-fragments">
                     {uncertainFragments.map((fragment, index) => (
                       <li key={index}>
-                        <FragmentariumLink item={fragment} />
+                        <FragmentariumLink
+                          item={{
```
fsimonjetz, line 180: Would it be more readable if we extract it to `const item = ...` and then call `<FragmentariumLink item={item} />`?

yCobanoglu: `uncertainFragment` instead `fragment`

yCobanoglu:
As discussed we get rid of uncertainFragments property and make the new behaviour as the default (not checking for isInFragmentarium only checking in frontend when necessary).

PR #281, opened 2022-07-14
Add old line numbers

fsimonjetz:
Implements [Add property “Old line number” to reconstructed lines](https://trello.com/c/kfscPvSj)


khoidt (changes requested):
Would make sense to refactor to solve the remaining two codeclimate issues. Otherwise, great job. :)

PR #287, opened 2022-08-11
Add omittedWords to WordInfo popover

fsimonjetz:
Implements omitted words part of [Tweak WordInfo popover](https://trello.com/c/Q27h9cQd)


src/corpus/domain/LineVariant.test.ts
```diff
@@ -1,85 +1,22 @@
+import {
```
fsimonjetz, line 1: The test tokens were all very similar. I generalized them into a number of test cases and extracted them into a separate file.

src/transliteration/ui/LineGroup.ts
```diff
@@ -12,9 +12,24 @@ export interface LineInfo {
   textService: TextService
 }
 
+function createLineTokens(
```
fsimonjetz, line 15: Extracts aligned tokens, including the omitted ones. The order of the result doesn't matter as it's sorted later on.

src/transliteration/ui/line-tokens.tsx
```diff
@@ -65,6 +65,7 @@ export class LineToken {
   token: LemmatizableToken
   lemma: DictionaryWord[] | null = null
   siglum: string | null = null
+  type = 'LineToken'
```
khoidt, line 68: `type: string = 'LineToken'`

src/transliteration/ui/LineLemmasContext.ts
```diff
@@ -1,6 +1,6 @@
 import React, { useContext } from 'react'
 import DictionaryWord from 'dictionary/domain/Word'
-import { LineToken } from './line-tokens'
+import { OneOfLineToken } from './line-tokens'
```
khoidt, line 3: Better use full paths, i.e. `'transliteration/ui/LineLemmasContext/line-tokens'`.

khoidt (approved):

src/transliteration/ui/line-tokens.tsx
fsimonjetz: Normally yet, but in this case the linter complains `Type string trivially inferred from a string literal, remove type annotation`.

src/transliteration/ui/LineLemmasContext.ts
fsimonjetz: That was an auto-import for sure. If we prefer absolute imports, maybe we should update the config so auto-imports are absolute by default.

PR #289, opened 2022-08-15
Modify how to cite

khoidt:


src/corpus/domain/chapter.ts
```diff
@@ -159,16 +159,33 @@ export class ChapterDisplay {
     return this.textHasDoi ? textIdToDoiString(this.id.textId) : ''
   }
 
+  mapNames(
+    names: readonly Author[] | readonly Translator[]
+  ): { family: string[]; given: string[] }[] {
+    return names.map((_name) => {
```
fsimonjetz, line 165: Is there a reason for using the `_` prefix? AFAIK, per convention that indicates unused parameters (and suppresses the "unused parameter" error), but that doesn't apply here.

fsimonjetz: Come to think of it, the `Author` and `Translator` interfaces are quite similar, and the logic here is specific to them. Maybe it would be worth it to change them into classes and add `family` and `given` getters, as well as the `nameToString` function below as method?

src/corpus/domain/chapter.ts
```diff
@@ -159,16 +159,33 @@ export class ChapterDisplay {
...
+      return {
+        family: _name.name,
+        given: _name.prefix,
+      }
+    })
+  }
+
+  mapAuthorsByRole(role: string): { family: string[]; given: string[] }[] {
```
fsimonjetz, line 173: I'd call it `getAuthorsByRole`, I think it would make the intention even clearer! :)

src/corpus/ui/HowToCite.tsx
```diff
@@ -45,18 +44,83 @@ function ExportButton({
...
 
+function nameToString(name, format = 'Name, GivenName', initials = false) {
+  const givenName =
+    initials === true
+      ? name.given
+          .split(/[\s,-]+/)
+          .map((n) => `${n[0]}.`)
+          .join(' ')
+      : name.given
+  console.log(givenName, name.given)
```
fsimonjetz, line 55: Remove debugging log

src/corpus/ui/HowToCite.tsx
```diff
@@ -45,18 +44,83 @@ function ExportButton({
   )
 }
 
+function nameToString(name, format = 'Name, GivenName', initials = false) {
```
fsimonjetz, line 47: `name` should have a type

src/corpus/ui/HowToCite.tsx
```diff
@@ -45,18 +44,83 @@ function ExportButton({
...
+  console.log(givenName, name.given)
+  return format === 'Name, GivenName'
+    ? `${name.family}, ${givenName}`
+    : format === 'GivenName Name'
+    ? `${givenName} ${name.family}`
+    : `${givenName}, ${name.family}`
+}
+
+function namesToString(
+  names: any[],
```
fsimonjetz, line 64: `names` should have a more specific type if possible

src/corpus/ui/HowToCite.tsx
```diff
@@ -45,18 +44,83 @@ function ExportButton({
...
+  prefix = '',
```
fsimonjetz, line 65: I wonder if it would simplify things if we remove the `prefix` parameter (since you only seem to use it once to add "Translated by") and instead add some logic in `CitationText` to check if there are translators, and if so, add the prefix.

khoidt: It is used twice, also in the 'With contributions by' case, so I'm not sure if it would simplify things.

fsimonjetz (approved):

khoidt:
[Modifications to the How to Cite dialogue](https://trello.com/c/Rhisb5uB/621-modifications-to-the-how-to-cite-dialogue)

PR #290, opened 2022-08-19
Modify display of Arabic words

ejimsan:


ejimsan:
[Modify Display of Arabic Words](https://trello.com/c/zaef4ViE/629-modify-display-of-arabic-words)

PR #297, opened 2022-10-08
Cosmetic changes to corpus search

ejimsan:


ejimsan:
[Display mode for Corpus Search](https://trello.com/c/RoMVkA7H/608-display-mode-for-corpus-search)

PR #304, opened 2022-10-31
Fix lemma popover

fsimonjetz:
Prevent `LineLemmasContext` from searching for empty lemmas


fsimonjetz:
[Fix omitted word in variant breaking lemma popover](https://trello.com/c/P9KCLTSQ/641-fix-omitted-word-in-variant-breaking-lemma-popover)

fsimonjetz:
[Fix omitted word in variant breaking lemma popover](https://trello.com/c/P9KCLTSQ/641-fix-omitted-word-in-variant-breaking-lemma-popover)

PR #313, opened 2022-11-23
Dictionary Fragmentarium Lemma Search

fsimonjetz:


src/query/QueryResult.ts
```diff
@@ -0,0 +1,16 @@
+import MuseumNumber from 'fragmentarium/domain/MuseumNumber'
+
+type LemmaSequence = ReadonlyArray<ReadonlyArray<string>>
+
+export interface QueryItem {
+  id_: string
+  museumNumber: MuseumNumber
+  matchingLines: readonly number[]
+  matchCount: number
+  lemmaSequences?: readonly LemmaSequence[]
```
fsimonjetz, line 10: Do we need this in frontend?

src/dictionary/ui/display/WordDisplay.test.tsx
```diff
@@ -218,7 +230,9 @@ let container: HTMLElement
 describe('Fetch word', () => {
   beforeEach(async () => {
     const genres = ['L', 'D', 'Lex', 'Med']
+    const queryResult = { items: [], matchCountTotal: 0 }
```
fsimonjetz, line 233: Add some data

src/query/QueryRepository.test.ts
```diff
@@ -0,0 +1,56 @@
+import { ApiQueryRepository } from './QueryRepository'
+import { testDelegation, TestData } from 'test-support/utils'
+import { QueryResult } from './QueryResult'
+
+const apiClient = {
+  fetchJson: jest.fn(),
+  postJson: jest.fn(),
+}
+const queryRepository = new ApiQueryRepository(apiClient)
+const lemmas = 'foo I+bar II'
```
fsimonjetz, line 10: maybe use comma instead of plus?

src/query/QueryResult.ts
fsimonjetz: It's not needed, not even in the backend (only used by MongoDB) so I'll remove it.

src/query/QueryRepository.test.ts
fsimonjetz: Won't fix for now

PR #350, opened 2023-02-27
Add Fragmentarium search tab hash and fix chapter line links

fsimonjetz:
- Adding a `#corpus` anchor to links to the Fragmentarium search now opens the Corpus tab and clicking on the tabs updates the hash to easier share links; if no hash is given, the Fragmentarium tab is opened by default
-  Clicking on line numbers in the Chapter view now updates the url and scrolls to the target line without reloading the page

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -241,6 +241,7 @@ function App({
                   fragmentQuery={parseFragmentSearchParams(location)}
                   wordService={wordService}
                   textService={textService}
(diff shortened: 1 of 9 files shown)


fsimonjetz:
[Don't reload chapter when clicking line number anchor](https://trello.com/c/RPI40ybL/669-dont-reload-chapter-when-clicking-line-number-anchor)

PR #391, opened 2023-09-08
Show Variant Label in Partial Search Results

fsimonjetz:
When only the 2nd variant matches, e.g., in https://www.ebl.lmu.de/fragmentarium/search/?transliteration=ka-li#corpus in "Literature > Poem of Creation (Enūma eliš) > Standard Babylonian I", line 61, it would display it as if it was the reconstruction, which is confusing. With this change, it shows the variant label in addition to the line number in such cases.

### DIFF
--- a/src/corpus/ui/ChapterView.sass
+++ b/src/corpus/ui/ChapterView.sass
@@ -9,6 +9,9 @@
         & ~ &
             padding-left: 2em
 
+        .chapter-display__variant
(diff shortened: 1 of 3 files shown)


fsimonjetz:
[Redesign Partial Chapter Display](https://trello.com/c/dz8LN5RV/714-redesign-partial-chapter-display)

PR #398, opened 2023-10-25
AfO Register

khoidt:


khoidt:
[Add AfO Register](https://trello.com/c/G3mhvOfq/717-add-afo-register)

src/afo-register/domain/Record.ts
```diff
@@ -0,0 +1,94 @@
...
+    this.text = text
+    this.textNumber = textNumber
+    this.linesDiscussed = linesDiscussed
+    this.discussedBy = discussedBy
+    this.discussedByNotes = discussedByNotes
+    this.fragmentNumbers = fragmentNumbers
+  }
+
+  toMarkdownString(): string {
+    let result = this.text + (this.textNumber ? ' ' + this.textNumber : '')
```
fsimonjetz, line 65: I recently made a little helper function `padLeft` in src/fragmentarium/domain/archaeology.ts because I had to do the `x ? ' ' + x : ''` thing over and over, which is then just `padLeft(this.textNumber)`. If you like you can move it to common/utils.ts or something and use it here, it might lower the complexity a bit.

PR #401, opened 2023-10-31
Add Metropolitan Number to Display

zsomborfoldi:


src/fragmentarium/ui/info/ExternalResources.test.tsx
```diff
@@ -61,6 +63,11 @@ describe('external resources', () => {
       'https://doi.org/10.11588/heidicon/',
       hilprechtHeidelbergNumber,
     ],
+    [
+      'The Metropolitan Museum of Art',
```
fsimonjetz, line 67: Also add `'The Metropolitan Museum of Art'` below in the `test.each` block starting at line 98 to ensure the label is not shown when the number is missing.

PR #404, opened 2023-11-03
Findspot display

fsimonjetz:
Add a way to display and select findspots in the Archaeology tab


fsimonjetz:
[Add Findspot to Archaeology](https://trello.com/c/M2SdSeCU/46-add-findspot-to-archaeology)

PR #410, opened 2023-11-23
Show suggestion when entering wrong number format in search

fsimonjetz:


fsimonjetz:
[Add validation for id search](https://trello.com/c/LJBYoDnk/728-add-validation-for-id-search)

PR #421, opened 2023-12-05
Update latest additions

fsimonjetz:
update route and make result list responsive


fsimonjetz:
[Redesign latest additions list](https://trello.com/c/vJh2S2Cu/730-redesign-latest-additions-list)

PR #463, opened 2024-03-11
Filter Indirect Joins

fsimonjetz:
Implements the `excludeIndirectJoins` function that filters manuscripts passed to the `Manuscripts` component for display such that indirect joins are excluded. They should only be displayed if 
1. they contain the manuscript museum number they are associated with or
2. they are not listed anywhere else in the group.

1. is trivial but 2. requires all join groups to be collected in order to check for uniqueness.


fsimonjetz:
[Filter indirect joins in Corpus Manuscript display](https://trello.com/c/IHvYwVP9/750-filter-indirect-joins-in-corpus-manuscript-display)

PR #469, opened 2024-04-22
added function to fetch following or preceding signs in requested order

Vas9ka:


src/signs/domain/Sign.ts
```diff
@@ -10,7 +10,17 @@ export interface Logogram {
   schrammLogogramme: string
   unicode: string
 }
+export class OrderedSign {
```
fsimonjetz, line 13: `OrderedSign` can probably just be an interface.

PR #483, opened 2024-06-14
Feature/cuneiform converter

Vas9ka:
*Added cuneiform converter
*Added more forms


Vas9ka (commented):
* Added Cuneiform Converter to /Tools/CuneiformConverter

PR #689, opened 2026-02-24
dev: moved provenance handling to DB-backed API data 

Fabdulla1:
This PR replaced frontend hardcoded provenance usage with DB-backed API data. It adds provenance endpoint support (/provenance, /provenances/{id}, /provenances/{id}/children), introduces cached provenance fetching in the service layer, and refactors provenance consumers to use dynamic data. 

It also hardens behavior for loading, empty responses, API failures, unknown provenance values, coordinates presence/absence, and parent/children provenance handling, including in-place enrichment of fallback records once API data arrives. 

Tests and mocks were updated to use API shaped provenance responses, and targeted suites for service, search, archaeology, colophon, manuscript, and dossier provenance behavior pass.


khoidt (changes requested):

khoidt (changes requested):
There are a few issues that should be addressed -- please check my comments. Other than that looks good!

Please don't merge before the backend is ready -- I still have to see how both actually work together.

khoidt (approved):
I think it's ready!

PR #722, opened 2026-05-06
dev: lazy route module loading and chunk isolation

Fabdulla1:
This PR implements a lazy loading of non-critical route modules so they don't inflate the initial application bundle. Home and About routes stay eager. 

**Affected Routes**:
tools, signs, bibliography, dictionary, corpus, library, projects, footer legal pages, and sitemap

**Additional Changes**:

- Added one top level suspense boundary around runtime route rendering with a dedicated route-loading fallback. 
- Introduced a shared sitemap config/types module andmoved sitemap defaults and slug types there.
- Decoupled sitemap generation route composition from the runtime router to avoid pulling all route modules into the runtime import graph
- Updated route modules and related type imports to use the shared sitemap config/types
- Added varying tests


khoidt (changes requested):
# PR #722 Review Summary

Hi! Here's a quick overview of the PR #722 audit. The full details are in [TASK-722-review.md](https://github.com/user-attachments/files/28012826/TASK-722-review.md)

- `yarn lint` and `yarn tsc` both pass with zero errors ✓
- Route path patterns (`/tools/*` etc.) verified to correctly match the RR v6 compat layer ✓
- `sitemapConfig.ts` extraction is a clean improvement to import structure ✓
- The `NotFoundPage` for `/projects/*` was a real missing guard ✓
- New `router.lazy-loading.test.tsx` validates the core lazy isolation concept ✓

## Concerns

There are **4 blockers** and 8 lower-priority issues. Main themes:

1. **Code duplication introduced, not removed.** The old `WebsiteRoutes` export was removed, but `sitemap.tsx` immediately grew its own private copy with nearly identical logic. The PR trades one shared function for two diverged functions.

2. **Four places must be kept in sync to add a new route group.** There are now two arrays and two maps enumerating the same route groups across three files. TypeScript won't enforce consistency between all of them.

3. **Lazy loading tests cover 5 of 9 route groups.** Signs, bibliography, dictionary, and corpus have no lazy isolation assertions — a future regression in those groups would be undetected.

4. **`AppDriver.waitForText` changed globally.** All existing integration tests now implicitly wait for the Suspense spinner to disappear before looking for text. This is a breaking change to the shared test driver that affects every caller.

khoidt (approved):
The code is good — lazy loading works, no bugs, tsc/lint/tests all green and console-clean. All 4 previous blockers are fixed.

Remaining (all minor):
- **R-1** — `SlugsArray` type is declared identically in `sitemapConfig.ts:7` and `sitemap.tsx:32`.
  Export it from `sitemapConfig.ts` and import it in `sitemap.tsx`, then delete the copy. This is
  the one item worth doing pre-merge, since the repo treats DRY as a hard gate.
- **R-2** — the top-level route `Suspense` fallback (`router.tsx:55`) borrows `withData-spinner`,
  a class meant for in-content data spinners. If that class's CSS is ever tweaked, the route
  loader changes too. Give it its own class, e.g. `route-loading-spinner`.
- **R-3** — the `Introduction` `<Route>` is built separately in `router.tsx:93` and
  `sitemap.tsx:57` (sitemap only adds `{...(sitemap && sitemapDefaults)}`). The rest of the
  composition is already shared via `composeWebsiteRoutes`; this last bit could be too.
- **R-4** — `sitemap.test.tsx:128` reads `saveAsMock.mock.calls[1][0]`, assuming the 2nd `saveAs`
  call is the chunk holding `/projects/CAIC`. If chunking changes it silently tests the wrong
  blob. Find the chunk by scanning the calls for one that contains `/projects/CAIC`.
- **R-5** — two intentional but unexplained choices: `about` stays eager (excluded from lazy
  loading in `websiteRouteGroups.ts:107`) and `footer` is registered with two `exact` paths
  instead of a wildcard (`websiteRouteGroups.ts:89-103`) so unknown legal subpaths reach the
  global not-found without loading the footer chunk. A one-line comment on each prevents a
  future "cleanup" from breaking the intent.
- **R-6** — `App.test.ts:62` allows 15 s for the parametrised render test. Plausibly justified by
  Suspense resolution across many routes in jsdom (suite ran in ~8 s locally), but a large fixed
  timeout can hide future slowdowns. Reduce it if the lazy imports can be pre-resolved/mocked there.
- **R-7** — qlty flagged ~21 duplicated lines across 4 spots in `notFoundRoutes.test.tsx`
  (~line 443). Test scaffolding only; extract a helper if you want to silence it.

Please fix before merging.

PR #723, opened 2026-05-07
CI: Add bundle size tracking with baseline bootstrapping

Fabdulla1:
This is PR 2 that is stacked on top of map_phase1.1 and is intended to be reviewed relative to **branch** map_phase1.1.

**Purpose**:
This change adds automated bundle size tracking before upcoming map integration in order to track heavy dependencies at initial app bundle. This will help track bundle growth as well as help in preventing unwanted load times for users who are not interacting with the planned map. 

**Changes**:

- Added dedicated bundle-size CI job in main workflow
- Added a bundle analysis and budget tooling in `bundleSizeMetrics.mjs`, `runSourceMapExplorer.mjs`, `checkBundleBudjet.mjs`, `checkBundleBudgetCli.mjs`, `recordBundleBaseline.mjs`, `bundle-size-baseline.json`
- Added package script for local and CI usage in `package.json`
- Added small documentation in `README.md` of local usage`. 
- Added module declaration support for mjs imports in `declarations.d.ts`
- Added necessary tests

**Updated CI Behavior**:
- Build with source maps and run source-map-explorer.
- Generate a bundle budget report
- Enforce 10% threshold on initial js bytes

**Notes**:
- Decided to use mjs file to avoid eslint error that prevented require() usage.


khoidt (changes requested):
The bundle-size tracking idea is well-structured and the lazy-loading refactor is a meaningful performance improvement. That said, there are several things that need attention before this is ready to merge.

Review:
[TASK-723-review.md](https://github.com/user-attachments/files/28006895/TASK-723-review.md)

# Summary:

## Hard blockers

- **ESLint error in `checkBundleBudget.mjs`** — `yarn lint` only scans `src/` so it passed silently, but running ESLint directly on `scripts/` reveals a real formatting error. The lint script needs to cover `scripts/**/*.mjs` too.
- **Test coverage at 44 %** — the main entry point (`runBundleBudgetCheck`), the write/log helpers, and several error paths have zero tests. The project aims to having 100 %.

## Things that need fixing

- The test file for the Node CI scripts landed in `src/common/utils/` (a browser app folder), which also forced a global `declare module '*.mjs'` type declaration just to compile. It should live next to the scripts it tests.
- `validateBaseline` is 46 lines with the same type-check + range-check guard written three times. It also silently injects a default value mid-function — validation and default-injection should be separate concerns.
- `createBundleBudgetReport` is 48 lines mixing delta maths and report assembly — it needs to be decomposed.
- `spurlockLink` should be `SpurlockLink` — React component names must be PascalCase.
- `WebsiteRoutes` (`sitemap.tsx`) and `RuntimeWebsiteRoutes` (`router.tsx`) are almost identical. The shared logic and route-module registry should live in one place.
- The `source-map-explorer` binary path is hard-coded into `node_modules/` — that's fragile.
- The CI `yarn install` step passes a `--dev` flag that Yarn v1 doesn't document for `install`. It's harmless but misleading.
- The lazy-loading test covers only 5 of 9 route groups; `signs`, `bibliography`, `dictionary`, and `corpus` are untested.

## Nice-to-have

- The PR bundles three unrelated concerns (CI tooling, router refactor, museum data) — separate PRs would make rollbacks and review much easier.
- Snapshot diffs contain factory-randomised data that changes on every fixture shape change — those snapshots aren't providing real regression value.

PR #729, opened 2026-05-20
Caching layer

Fabdulla1:
Adding caching layer:
- Added TTL + bounded caches in FragmentService
- Added in-flight request deduplication for cacheable reads
- Added safe handling to avoid caching stale responses when an older in-flight request resolves after newer state.
- Added cache invalidation on mutating operations so reads & queries do not serve stale data
- Added cache-scope awareness and full cache reset when scope changes
- Wired cache scope from InjectedApp so authenticated users and guests do not share cached data.


khoidt (changes requested):
Hi, this generally looks good, but some changes are still required. Details here:

[TASK-729-review.md](https://github.com/user-attachments/files/28150517/TASK-729-review.md)

PR #733, opened 2026-05-26
Auto-create `.env.local` and inject Codespaces secrets

khoidt:
Implement automatic creation of `.env.local` from `.env.test` and inject matching Codespaces secrets during the development container setup. Update documentation to reflect these changes and provide clear instructions for users.


.devcontainer/init.sh
```diff
@@ -0,0 +1,41 @@
...
+            flags=re.MULTILINE
+        )
+        injected.append(key)
+
+if injected:
+    with open('.env.local', 'w') as f:
+        f.write(content)
+    print('Injected Codespaces secrets into .env.local: ' + ', '.join(injected))
+else:
+    print('No Codespaces secrets found — .env.local uses placeholder values from .env.test')
```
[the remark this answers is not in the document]
khoidt: Fixed: init.sh now only injects a secret when the current .env.local value still matches the placeholder from .env.test. Manual local edits are preserved on subsequent initializeCommand runs.

.devcontainer/init.sh
```diff
@@ -0,0 +1,41 @@
...
+with open('.env.local') as f:
+    content = f.read()
+
+injected = []
+for key in keys:
+    value = os.environ.get(key, '')
+    if value:
+        content = re.sub(
+            r'^' + re.escape(key) + r'=.*',
+            key + '=' + value,
```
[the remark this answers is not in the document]
khoidt: Fixed: replaced the Python regex substitution flow with shell/awk line replacement, so secret values are written literally and backslashes are not interpreted as replacement escapes.

.devcontainer/devcontainer.json
```diff
@@ -20,6 +20,7 @@
     }
   },
 
+  "initializeCommand": "bash .devcontainer/init.sh",
```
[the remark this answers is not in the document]
khoidt: Fixed: init.sh no longer depends on python3 on the host. The script is now shell+awk only, so initializeCommand works without requiring host Python from the Docker image build.

README.md
```diff
@@ -35,7 +35,12 @@ This project is configured to work with [GitHub Codespaces](https://github.com/f
 3. Create a new codespace or open an existing one
 4. **Important:** For dev containers to work properly, open the codespace in the **desktop version of VS Code** rather than the browser. When the codespace starts, VS Code will prompt you to open it in the desktop application.
 5. Once opened in VS Code desktop, the dev container will automatically set up the development environment
-6. Configure the required environment variables in `.env.local` (see [Running the application](#running-the-application) section)
+6. `.env.local` is created automatically from `.env.test` before the container
+   builds. If you have configured [Codespaces secrets](https://docs.github.com/en/codespaces/managing-your-codespaces/managing-secrets-for-your-codespaces)
+   with the same names as the keys in `.env.test`, they are injected into
+   `.env.local` automatically and the app will be fully configured on first start.
+   Otherwise, open `.env.local` inside the container and fill in your credentials
+   (see [Running the application](#running-the-application) section).
```
Fabdulla1: So for many files, /.devcontainer/devcontainer.json, b/.devcontainer/init.sh at lines 23, 21, 19-24, the implementations rely on the `initializeCommand` plus printenv to read Codespaces secrets before container build. But, the automatic secret injection into .env.local on first start still fails in Codespaces. I think if you move secret resolution to a post-create/post-start lifecycle step that runs inside the running container it might be better.

.devcontainer/init.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
+set -e
+
+ensure_env_local_permissions() {
+    chmod u+rw,go+r .env.local 2>/dev/null || sudo chmod u+rw,go+r .env.local
```
Fabdulla1: .env.local permissions are forcibly relaxed to `u+rw,go+r` and increase exposure on shared hosts/dev environments. Maybe we can make permissions stay least-privilege and ownership/access mimsatches are handled without broadening file readability.

Also, fallback to sudo inside the initialize flow can fail or block non-interactive contexts. If chmod fails and sudo is unavailable or prompts for a password the set up will abort unexpectedly. We could handle permission failures explicitly with a clear message and avoid interactive privilege escalation in lifecycle scripts

.devcontainer/devcontainer.json
Fabdulla1: Starting up failed for me and it's because the host command here is hardcoded to bash. Users without bash in PATH will fail before container creation even if Docker/Dev containers are otherwise configured. It would be good to have shell-agnostic initialize invocation or maybe you could move this logic into an in-container lifecycle hook so host shell differences don't break setup

.devcontainer/inject-secrets.sh
```diff
@@ -0,0 +1,63 @@
...
+            secret_value=$(printenv "$key" || true)
+
+            if [ -z "$secret_value" ]; then
+                continue
+            fi
+
+            current_line=$(grep -m1 "^${key}=" .env.local || true)
+            if [ -z "$current_line" ]; then
+                continue
+            fi
```
Fabdulla1, line 42: If .env.local exists but misses a key newly added to .env.test it seems the script will silently skip that key. Devs could end up with partially configured environments after new vars are introduced with no signal about missing keys. Could add logic for when key exists in .env.test but is missing in .env.local then we could append it so devs don't miss required configs after updates

Fabdulla1 (changes requested):

Fabdulla1 (approved):

PR #743, opened 2026-06-04
Add Dictionary of Realia tool

khoidt:
Introduce the Dictionary of Realia tool, including authentication scopes, domain types, repository, and service. Implement a search UI with results display and detail pages. Conduct a security audit and address findings. Ensure comprehensive test coverage across new components.


src/realia/infrastructure/RealiaRepository.ts
Fabdulla1, line 1: I noticed that `/realia?query=${query}` could have reserved characters like "pig & cow" sent as `/realia?query=pig & cow" where `& cow` could be interpreted as a separate query parameter. I think this could also affect searches with spaces, Unicode and etc. Maybe we change this to encode as we have in the codebase and add tests to harden behavior?

Fabdulla1 (changes requested):
Just one comment, other than that, everything looks great :D

src/realia/infrastructure/RealiaRepository.ts
khoidt: Good catch — fixed in 563bd213: `search` now encodes the query with `encodeURIComponent`, matching the existing convention (e.g. `WordRepository`). Added table-driven tests asserting the exact encoded URL for reserved characters (`pig & cow` → `query=pig%20%26%20cow`), spaces, Unicode (`Ninĝirsu`), and a reserved-character mix. Also verified against a gunicorn+Falcon instance using the same `req.get_param("query")` handler that the encoded value arrives intact, so no API change is needed.

PR #748, opened 2026-06-15
dev: added optimization & tests

Fabdulla1:
## Summary

- Render fragment search results from backend-provided summary data, avoiding a separate fragment request for every result.
- Preserve compatibility with legacy query items by falling back to per-fragment hydration when no summary fragment is present.
- Fix cancelled in-flight cache reuse and stabilize `withData` watch keys.
- Add coverage for summary-backed search results and compatibility paths.

## Backend contract

This frontend change depends on fragment query responses optionally returning summary fields including `description`, `script`, `hasPhoto`, `matchingLinePreview`, and `thumbnailPath`. The backend contract does not currently provide an explicit `type`/`kind` discriminator, so the frontend uses those required summary fields for structural detection while retaining the legacy response path.

### DIFF
--- a/src/common/utils/getOrFetchCachedValue.ts
+++ b/src/common/utils/getOrFetchCachedValue.ts
@@ -29,7 +29,11 @@ export default function getOrFetchCachedValue<CacheKey, CacheValue>({
   const inFlightRequest = requests.get(key)
 
   if (inFlightRequest) {
-    return inFlightRequest
+    if (inFlightRequest.isCancelled()) {
+      requests.delete(key)
+    } else {
+      return inFlightRequest
+    }
   }
 
   const requestReference: { current?: Bluebird<CacheValue> } = {}
--- a/src/corpus/ui/search/CorpusSearchResult.tsx
+++ b/src/corpus/ui/search/CorpusSearchResult.tsx
@@ -181,15 +181,21 @@ export const CorpusSearchResult = withData<
 >(
   ({ data, textService, corpusQuery }): JSX.Element => {
     const chapterCount = data.items.length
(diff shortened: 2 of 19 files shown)


khoidt (approved):

khoidt (approved):
Solid, well-scoped PR. The search results page now renders straight from backend-provided
summary data instead of firing one fragment fetch per result — a real performance win — and it
falls back to the old per-result fetch when no fragment is present, so it's backward compatible.
It also slips in two genuine bug fixes (cancelled cached-promise reuse, and `withData` watch
identity). `tsc`, `lint`, and 278 tests across the 7 changed suites all pass console-clean.

No blockers. No correctness defects found. Findings below are minor/quality only.

Remaining (all minor):

- **F-1** — qlty flagged ~18 duplicated lines in `FragmentariumSearch.test.tsx` (~lines 259/281);
  test scaffolding, extract a helper if you want to silence it.
- **F-2** — `isQuerySummaryItemDto` discriminates by `'description'/'script'/'hasPhoto' in dto`;
  a backend `type`/`kind` discriminant would be sturdier.
- **F-3** — `createQuerySummaryFragment` hard-codes placeholders (`Museums.HYPERURANION`, empty
  `record`/`folios`/`publication`…); safe today because the search view doesn't render them, but
  worth a one-line "display-only lightweight fragment" comment so it isn't mistaken for complete.
- **F-4** — the new `watch` keys (`stringify(fragmentQuery)`, `dossiers…join('|')`) lean on
  serialization stability; fine in practice, negligible collision risk.
- **F-5** — PR has no description; add a short note on the backend summary-item contract it
  depends on (it's coupled to a backend change).

PR #750, opened 2026-06-16
dev: added map MVP

Fabdulla1:
This PR adds a frontend-only `/tools/map` tab that introduces a lightweight interactive map for exploring eBL findspots. The implementation uses MapLibre with existing frontend-accessible provenance data, converts provenances into GeoJSON, and renders clustered findspots with popups and deep links to fragment search.


khoidt (changes requested):
Good feature (the map looks fantastic!), clean code, all tests green, and the map works in the browser. Blocked only on two hard gates:

- 250-line ceiling — MapTab.test.tsx (415) and Tools.test.tsx (331) need splitting.
- 100% coverage — useFindspotMap.ts at 86%; the mouse-cursor handler is untested.

Also worth fixing: no error handler for the basemap CDN (silent blank map if it fails), and the map is mouse-only.

khoidt (changes requested):
Nice work! All four bot findings are properly fixed rather than waved through, and two of the three blockers from my previous review are closed. `yarn lint`, `yarn tsc` and qlty are all clean. However, there are still some issues to address.

### Blockers

- **F1 — the full test suite doesn't pass.** `yarn test --watchAll=false` exits 1:
  `MapTab.test.tsx`, `MapTab.interactions.test.tsx`, `MapTab.clickGuards.test.tsx` and `MapTab.errors.test.tsx` never load — `maplibre-gl/dist/maplibre-gl.css` (imported at `MapTab.tsx:3`) gets parsed as JS. Reproduced 3×, including the exact CI command with `CI=true`; it passes when run scoped, which is why CI and the last review missed it. Cause: `package.json`'s `transformIgnorePatterns` ignores *every* file under `node_modules` (CRA's default only ignores
  `.js/.mjs/.jsx/.ts/.tsx`), so that stylesheet can never be transformed — it's just the first `node_modules` CSS a test has ever reached. The `jest.mock(…, { virtual: true })` calls at `MapTab.testHelpers.tsx:67` and `:126` are papering over it, and `virtual: true` isn't for modules that actually exist: narrow `transformIgnorePatterns` to JS-ish files and the real MapLibre loads straight through the mock — 26 failures with `Failed to initialize WebGL`. Fix: add `"^maplibre-gl/dist/maplibre-gl\\.css$": "identity-obj-proxy"` to `moduleNameMapper` (already installed), drop both `virtual: true` flags, and move the `maplibre-gl` factory to `src/__mocks__/maplibre-gl.ts`. I verified the `moduleNameMapper` line alone, with no source changes: 347/347 suites, 3507 passed, exit 0, no console noise.
- **F2 — `useFindspotMap.ts` is 271 lines**, over the 250 ceiling — the same gate that blocked `MapTab.test.tsx` (415) and `Tools.test.tsx` (331) last round. Three clean extractions: popup parsing (`85–141`), cursor handling (`171–196`), basemap error classification (`198–211`). Leaves the hook around 90 lines.
- **F3 — coverage isn't 100%.** `createFindspotPopup.ts:59` — the `if (coordinates)` false path is never taken because every test supplies coordinates. 87.5% branch on that file, 98.93% across `src/map`. One test case with `coordinates: undefined` closes it.

### Medium

- **F4 — `makeProvenance` is duplicated verbatim** in `MapTab.testHelpers.tsx:129` and `provenanceToGeoJson.test.ts:5`. Understandable (importing the helper drags in its `jest.mock` side effects), but DRY is a hard gate — pull the fixture into a third, side-effect-free module both can import.

### Small stuff

- **F5** — `MAP_BACKGROUND_ERROR_PATTERNS` (`useFindspotMap.ts:198`): `MAP_STYLE_URL` can never be the deciding match (it already contains `basemaps.cartocdn.com` *and* `style.json`), and `'failed to fetch'` / `'failed to load'` will flag any
  tile, sprite or glyph failure as "the map background could not be loaded".
- **F6** — `openFindspotPopup` (`useFindspotMap.ts:143-152`) calls `getFeaturePointCoordinates` directly and again inside `getPopupProperties` (`:118`). Compute it once.
- **F7** — `useMapSourceData.ts:7` imports `fitMapToData` out of `useFindspotMap`, a hook module. Give it its own file (`mapBounds.ts`) — also helps F2.
- **F8** — `setCanvasCursor` (`useFindspotMap.ts:171-180`) has an inline comment (project rule: none unless asked) and a blanket `catch` around `map.getCanvas()`; the `canvas?.style` guard already covers everything the tests exercise, and nothing makes `getCanvas()` throw.
- **F9** — all 29 intra-module imports in `src/map` are relative (`./mapLayers`); the standard is `map/mapLayers`, which the rest of this diff follows. Fair caveat: ~510 relative imports already exist elsewhere and nothing lints them.
- **F10** — the popup says `Area boundary available` for polygons (`createFindspotPopup.ts:65`), but polygons are collapsed to a centroid (`provenanceToGeoJson.ts:43`) and no boundary is ever drawn. Reword or render it.
- **F11** — map interaction is still mouse-only (`useFindspotMap.ts:247-254`). The findspot link list is a fine text alternative for an MVP; just say so in the PR description so my point is closed rather than dropped.
- **F12** — `MapTab.tsx:53-58` sets state unconditionally when `fetchProvenances()` settles. Silent on React 18, but the tab is lazily mounted and easy to leave mid-flight; an `ignore` flag in the cleanup fixes it.
- **F13** — `MapTab.sass:31` declares `.maplibregl-popup-content a` at top level, so it goes document-wide once the chunk loads. Nest it under `.map-tab`.
- **F14** — `Tools.contentMocks.testSupport.tsx` registers its `jest.mock` calls as an import side effect, so it must be imported before `router/Tools`. It currently is, but nothing enforces it — an import reorder would silently pull the real pages into the tests.
- **F15** — branch is 5 commits behind `master` and `rebaseable: false`, so no CI run reflects the real merge.

khoidt (changes requested):
This is a really nice round. Thirteen of the fifteen findings from my last review are properly fixed, not papered over — the Jest `transformIgnorePatterns` diagnosis landed exactly as `moduleNameMapper` + a real `src/__mocks__/maplibre-gl.ts` with both `virtual: true` flags gone, `useFindspotMap.ts` came down from 271 to 179 lines through four genuine extractions, and the CodeQL URL check is now a proper `new URL()` host+path comparison instead of substring matching. I ran the app against a local provenance stub and the map is lovely: real CARTO tiles, clustering, pointer feedback, and popups that render `` as plain text with the link href percent-encoded. Zero console errors in the browser, and the full suite is 421/421 suites and 3983/3983 tests with no console output at all.

Four things block approval, and only one of them is really about the map code.

- **F16** — `yarn tsc` is red. It's pre-existing (it fails at the merge base too, arrived with #767), but our rules make it a hard gate and require fixing pre-existing failures rather than deferring them.
- **F17** — coverage isn't 100%. Exactly one statement in all of `src/map` never executes: the `return` inside `expandCluster`'s `.catch()`. It's unreachable-by-design dead code, and the test named for it can't fail either way.
- **F18** — F11 is still open. You just need a sentence in the PR description saying the findspot link list is the non-mouse alternative, and I'll close it.
- **F19** — F15 is still open. The branch is 5 commits behind `master`, so no green check reflects the actual merge result.

Beyond that: two DRY duplications that our own hard gate catches (F20, F21), one empty-state message that reads wrong when the backend simply has no data (F22, reproduced in the browser), and four small cleanups. Nothing here is deep — I'd expect all of it in one pass.

### Details

#### F16 — Blocker (pre-existing) — `yarn tsc` fails

`src/fragmentarium/ui/text-annotation/realiaOptionLoader.ts:30` — `error TS18046: 'entries' is of type 'unknown'.`

Not introduced by this PR. The file is untouched by the diff, and the error reproduces at the merge base `8971a666` — it arrived with #767. But our instructions treat `yarn tsc` as a hard gate and require pre-existing failures to be fixed at their root cause in the same task rather than deferred, so it can't ride along. (This is the same defect recorded as F14 in `TASK-765-review.md`, still unfixed on `master`.)

Root cause: `RealiaService.search` returns a Bluebird `Promise<readonly RealiaEntry[]>`, and `toNativePromise<T>(promise: PromiseLike<T>)` can't infer `T` from Bluebird's overloaded `then`, so `T` collapses to `unknown` and `entries` at line 30 is untyped.

Fix: bind the type argument explicitly — `toNativePromise<readonly RealiaEntry[]>(realiaService.search(query))` — or widen the parameter to accept Bluebird's promise type. Worth noting that no CI workflow runs `tsc`, which is exactly why this PR is green while the branch doesn't compile.

#### F17 — Blocker — coverage is not 100%; a no-op guard makes it unreachable

`src/map/useFindspotMap.ts:76-78`.

```js
    .catch(() => {
      if (!isActive()) return
    })
```

`src/map` reports **99.56% statements / 99.18% branches**, not 100%. I pulled the statement-level data out of `coverage-final.json` rather than trusting the summary column, and there is exactly **one** unexecuted statement in the whole directory: statement 19, line 77, columns 23–29 — the `return`. Its branch counts are `[0, 1]`: the `if` was evaluated once, the consequent never taken. This is stable, reproduced 3/3 on directory-scoped runs and again in the full suite, and it's what CI's `qlty coverage diff` of 99.0% is reporting.

The reason it can never be covered is that it can never matter. `return` is the last statement of the arrow body, so taking it and not taking it are the same thing — the handler does nothing in either direction. The only test that reaches this path (`useFindspotMap.test.tsx:98-118`) rejects while the component is still mounted, so `isActive()` is `true`; and even a mounted/unmounted variant would assert nothing, because the branch has no observable effect.

Fix: drop the `if` and leave the catch body empty — `.catch(() => undefined)` — or, better, decide what should actually happen when cluster expansion fails and do that. Silently swallowing a rejected `getClusterExpansionZoom` means a user clicking a cluster gets no zoom and no feedback. Either way, coverage returns to 100% without inventing a test for dead code.

#### F18 — Blocker — F11 was never closed

`useFindspotMap.ts` still registers `click`, `mousemove`, `mouseenter` and `mouseleave` only; there is no keyboard path to a findspot popup.

I said last round that the findspot link list is a fine text alternative for an MVP, and asked only that the PR description say so, so the point is closed rather than dropped. The PR description is still the original three-sentence version and doesn't mention accessibility at all. I verified the link list works in the browser — every provenance gets a real `/library/search?site=…` link — so this is purely a documentation ask.

Fix: one sentence in the PR description, e.g. "Map interaction is pointer-only in this MVP; the *Findspot searches* list below the map is the keyboard- and screen-reader-accessible equivalent, and links to the same fragment searches the popups do."

#### F19 — Blocker — F15 was never closed; the branch is still behind `master`

`git rev-list --left-right --count origin/master...HEAD` → `5 11`.

Five commits on `master` are not in this branch, so the green `test` / CodeQL / qlty checks on `037d71d7` describe a tree that isn't what would land. This matters more than usual right now: `master` has moved through #767 (which is what introduced the `tsc` failure in F16) and #781/#782 since your last merge.

Fix: merge `origin/master` (the branch has `rebaseable: false`, so merge rather than rebase) and let CI re-run.

#### F20 — Medium — DRY hard gate: point-coordinate extraction exists twice

`src/map/mapBounds.ts:8-21` (`getPointCoordinates`) and `src/map/findspotPopupProperties.ts:6-24` (`getFeaturePointCoordinates`).

Both do the identical thing: reject non-`Point` geometry, pull `coordinates[0]` / `coordinates[1]`, require both to be finite numbers, return `[longitude, latitude] | null`. One is written as a compound `&&` returning a ternary and the other as sequential `if` guards, but the domain rule — "what counts as a usable findspot coordinate" — is the same rule expressed twice. DRY is a hard gate in this project, and this is exactly the shape it's meant to catch: if the rule ever changes (say, a latitude range check), one copy will be missed.

Fix: keep one implementation — `getFeaturePointCoordinates` is the better-typed of the two — in its own small module (or export it from `findspotPopupProperties` and import it into `mapBounds`), have both call sites use it, and fold `mapBounds.test.ts`'s coordinate-rejection cases into the shared module's test.

#### F21 — Medium — DRY hard gate: 12 `jest.mock` blocks duplicated verbatim across two suites

`src/router/Tools.test.tsx:5-79` and `src/router/Tools.navigation.test.tsx:7-81`.

Splitting `Tools.test.tsx` was the right call and it fixed the 250-line blocker, but the split copied the entire mock preamble instead of sharing it. I checked with `diff <(grep -A2 '^jest.mock' Tools.test.tsx) <(grep -A2 '^jest.mock' Tools.navigation.test.tsx)` — the blocks are byte-identical, 12 in each file, about 62 lines duplicated. Adding a twelfth tool means editing both files, and if only one is updated the other silently pulls the real page component into the test.

I do appreciate why it looks unavoidable: `jest.mock` is hoisted, and F14 was specifically about a shared module registering mocks as an import side effect. But those two constraints don't force a copy-paste — a plain (non-hoisted) helper that both files invoke at the top, or a `jest.setup` module listed in `setupFilesAfterEach` for these suites, keeps the registration explicit while defining the list once. `Tools.contentMocks.testSupport.tsx` already holds the components; only the wiring is duplicated.

Fix: extract the module-path → mock-component mapping into one exported constant and drive both files' `jest.mock` calls from it, or move the shared registration into a setup module these two suites opt into.

#### F22 — Medium — the empty state says "no match" when there is simply no data

`src/map/MapTab.tsx:103-105`.

```jsx
{filteredProvenances && filteredProvenances.length === 0 ? (
  <Alert variant="info">No findspots match &ldquo;{filter}&rdquo;.</Alert>
) : null}
```

When the backend returns zero provenances, `filter` is `''`, so the user sees literally `No findspots match “”.` I reproduced this in the browser by pointing the app at a stub `/provenances` returning `[]` — the rendered alert text is exactly `No findspots match “”.` with the empty quotes visible. The existing test at `MapTab.test.tsx:123` ("does not crash with empty provenance data") only asserts the input and the map region exist, so it passes straight through this.

It's cosmetic, but it's the message a user sees on the day the endpoint returns nothing, which is precisely when you'd want it to be clear.

Fix: branch on whether a filter is active — `filter.trim() ? 'No findspots match “…”.' : 'No findspot locations are available.'` — and add the no-filter case to `MapTab.test.tsx`.

#### F23 — Low — `MAP_STYLE_URL` and its host/path constants can drift apart

`src/map/mapBackgroundError.ts:1-5`.

```ts
export const MAP_STYLE_URL =
  'https://basemaps.cartocdn.com/gl/positron-gl-style/style.json'

const MAP_STYLE_HOST = 'basemaps.cartocdn.com'
const MAP_STYLE_PATH = '/gl/positron-gl-style/style.json'
```

The same URL is written twice — once whole, once split. `MAP_STYLE_URL` is what the map actually loads; `MAP_STYLE_HOST` / `MAP_STYLE_PATH` are what `isStyleUrl` matches against. Change the basemap style and update only the first, and nothing breaks loudly: the map still loads from the new URL, but `isMapBackgroundLoadError` stops recognising its failures, so the "The interactive map could not be loaded" warning quietly never appears again. That's a silent failure of the exact feature this module exists to provide.

Fix: derive it — ``const MAP_STYLE_URL = `https://${MAP_STYLE_HOST}${MAP_STYLE_PATH}` `` — so the three can't disagree.

#### F24 — Low — 25 hand-written re-exports where `export *` would do

`src/map/MapTab.testHelpers.tsx:4-62`.

Lines 4–31 import 25 symbols from `map/testSupport/mapLibreMock` and lines 34–62 re-export the same 25 symbols unchanged. That's 58 lines doing what `export * from 'map/testSupport/mapLibreMock'` does in one, and it's the file's whole job apart from the two `FragmentService` factories at the bottom. `src/__mocks__/maplibre-gl.ts:1-2` already uses `export *` for the same module, so the pattern is established in this very PR. Practical cost: every new mock added to `mapLibreMock.ts` has to be added here twice more, and forgetting means a confusing "not exported" error in a test file.

Fix: replace lines 4–62 with `export * from 'map/testSupport/mapLibreMock'` plus the existing `makeProvenance` re-export, keeping `makeFragmentService` / `makeFailingFragmentService` as they are.

#### F25 — Low — the rejection test can't fail

`src/map/useFindspotMap.test.tsx:98-118`, "does not surface an unhandled rejection when expansion fails".

Its only assertion is `expect(mockEaseTo).not.toHaveBeenCalled()`, which holds whether the rejection was handled, unhandled, or never delivered — `easeTo` is only called on the *success* path, so a rejected promise trivially satisfies it. The thing the test is named for (that no unhandled rejection escapes) is never asserted. This is the same code as F17, so they'll be fixed together.

Fix: assert the actual contract — e.g. install an `unhandledRejection` listener for the duration of the test and assert it never fires, or await the settled promise and assert the handler ran. If F17 is resolved by deleting the no-op guard, this test should be rewritten to match whatever the catch is then meant to do.

#### F26 — Low — non-null assertion hides the hook's invariant

`src/map/useFindspotMap.ts:140` — `initializeFindspotSource(map, latestProvenancesRef.current!)`.

The assertion is correct today: the effect returns early unless `isReady`, and `isReady` is `provenances !== null`. But the invariant lives 14 lines away from the `!`, and the ref is deliberately mutable so a future change could make it null between registration and the `load` event. A `const provenances = latestProvenancesRef.current; if (!provenances) return` inside `handleLoad` costs two lines and removes the assertion.

#### F27 — Low — the link list includes findspots that can never appear on the map

`src/map/MapTab.tsx:113-129`.

The *Findspot searches* list is built from `filteredProvenances`, while the map features are built from `provenanceToGeoJson(filteredProvenances)`, which drops any provenance without usable geometry. In the browser, a provenance with no `coordinates` and no `polygonCoordinates` ("Placeless" in my stub) appears in the list under a map it has no marker on.

I think this is probably the behaviour you want — the link is a fragment search, and it works fine without coordinates — but it isn't stated anywhere and it isn't tested, so it reads as an oversight next to `MapTab.test.tsx:134`, which explicitly asserts geometry-less provenances are excluded from the map. A one-line test asserting the list *does* include them (or a filter, if the opposite was intended) would settle it.

khoidt (approved):
This is a really good round. Every one of the twelve findings I left last time is fixed at the root, plus all four bot findings, and I re-checked each against the code rather than trusting the commit messages. Two are fixed better than I asked: the `tsc` failure gets the exact type argument I suggested, and the duplicated `jest.mock` preamble wasn’t just deduplicated but *guarded* with an `expectToolsContentPagesMocked` assertion in each suite, which closes the older “nothing enforces the import order” point at the same time.

The `src/map` layering into `domain/` / `maplibre/` / `ui/` is a real improvement — the pure logic is now testable without MapLibre at all, and those tests are unusually thorough (the coordinate validator alone covers twelve rejection cases including transposed coordinates and `1e308`). The suite is 429 suites / 4151 tests green with zero console output.

One thing to flag before merge: this PR changes `.dockerignore`, and the `docker` / `docker-test` jobs only run on pushes to `master`, so nothing validates that change until after it lands. Details in the config warning below. Otherwise the only thing standing between this and merge is mechanical — the branch is one commit behind `master`, so no green check describes the tree that would actually land.

### Details

**N1 — Low — The suggestion dropdown and the map/list results disagree on whitespace.** [`FindspotFilterInput.tsx:54-56`](src/map/ui/FindspotFilterInput.tsx#L54-L56) vs [`findspotFilter.ts:11`](src/map/domain/findspotFilter.ts#L11). `filterProvenances` folds `filter.trim()`; the select’s custom `filterOption` folds the raw input. Supplying a custom `filterOption` bypasses react-select’s default `createFilter({ trim: true })`, so nothing trims it. Typing `bab` → dropdown and results both show *Babylon*; typing `bab ` (trailing space) → **dropdown empty, map and list still show Babylon**; same for a leading space; typing only spaces → dropdown empty while everything is unfiltered below. With `noOptionsMessage={() => null}` the mismatch is silent — the menu just goes blank. **Fix:** export one predicate from `findspotFilter.ts`, e.g. `matchesFindspot(name, query)` that trims and folds, and have both `filterProvenances` and `filterOption` call it. That also satisfies the DRY gate for the folding rule.

**N2 — Low — Sprite, glyph and style-parse failures now produce neither a banner nor a Sentry report.** [`mapBackgroundError.ts:49-70`](src/map/maplibre/mapBackgroundError.ts#L49-L70), [`useFindspotMap.ts:181-190`](src/map/maplibre/useFindspotMap.ts#L181-L190). `handleError` shows the banner only when `isMapBackgroundLoadError` matches the style JSON URL exactly, and otherwise reports only when `getReportableMapError` returns non-null, which requires the event to be source- or layer-scoped. An error that is neither — a sprite or glyph fetch failure, or a style parse error — falls through both and is discarded. This looks deliberate: `mapBackgroundError.test.ts` explicitly asserts `['a sprite failure', { error: { message: 'Not Found' } }] → null`. But a missing sprite means the map renders without its icons and a missing glyph without its labels, with no banner for the user and no event in Sentry — which is the one failure mode this module exists to make visible. **Fix:** report (without the banner) any error carrying a message that is not a per-tile miss, or add a third branch for sprite/glyph URLs.

**N3 — Low — DRY: the two error predicates repeat their unwrapping, and one disjunct is unreachable.** [`mapBackgroundError.ts:35-70`](src/map/maplibre/mapBackgroundError.ts#L35-L70). `isMapBackgroundLoadError` and `getReportableMapError` open with the same six lines — reject non-objects, pull `.error`, reject non-object errors. Extracting `toMapErrorEvent(event): MapLibreErrorEvent | null` removes the duplication and makes both read as one guard plus one rule. Separately, `getReportableMapError` requires `mapEvent.tile !== undefined` to be **false** and then calls `isSourceOrLayerScoped`, whose third disjunct is `event.tile !== undefined` — unreachable from that call site. The combined condition is correct but a reader has to work that out; splitting into `isSourceOrLayerScoped` (sourceId/layer only) and a separate `isTileScoped` would make both call sites say what they mean.

**N4 — Nit — The debounce in `useMapSourceData` is doing nothing the effect cleanup isn’t.** [`useMapSourceData.ts:26-31`](src/map/maplibre/useMapSourceData.ts#L26-L31) creates a fresh debounced function on every effect run and invokes it exactly once, returning `() => fitToFilteredData.cancel()`. lodash’s trailing-edge coalescing therefore never sees a second call — the coalescing comes entirely from the cleanup cancelling before the next run schedules its own. The behaviour is right, and the new “coalesces rapid filter changes into a single camera fit” test pins it, but `debounce` reads as if lodash were doing the work. **Fix:** say it directly with `setTimeout`/`clearTimeout`, or hoist the debounced function with `useMemo` so lodash genuinely does the coalescing.

**N5 — Nit — `foldForSearch` embeds literal invisible combining marks.** [`foldForSearch.ts:1`](src/map/domain/foldForSearch.ts#L1) — `const NON_SPACING_MARKS = /[̀-ͯ]/g` is a range of raw U+0300-U+036F combining characters. It renders as visual noise (or invisibly attaches to the surrounding brackets) in editors, diffs and review UIs, and a stray copy/paste can silently change the range. `/[\u0300-\u036f]/g` is identical and legible. (`NON_DECOMPOSING` is fine — those characters are visible.) The folding itself is correct: `Aššur → assur`, `Ḫattuša → hattusa`, `Ĝirsu → girsu`, `Tell ʿAqar → tell aqar`.

**N6 — Nit — Cluster colours bypass the design token this PR introduces.** [`mapLayers.ts:26-36`](src/map/maplibre/mapLayers.ts#L26-L36). `unclusteredLayer` correctly uses the new `EBL_COLOR_BRAND_PRIMARY` token, with `designTokens.test.ts` keeping it in sync with the Sass source — a good idea. The cluster step colours right above it are `#51bbd6` / `#f1f075` / `#f28cb1`, the values straight from MapLibre’s own clustering example, hard-coded. Contrast against the default black cluster-count label is fine, so this is cosmetic, but it leaves the map half tokenised.

**N7 — Low — A typed filter cannot be cleared with the clear button, and “no options” is silent.** [`FindspotFilterInput.tsx:33-36`](src/map/ui/FindspotFilterInput.tsx#L33-L36) and [`:63`](src/map/ui/FindspotFilterInput.tsx#L63). `value={selectedOption}` is non-null only when `filter` exactly equals a provenance’s `longName`. react-select renders the clear indicator only when it has a value, so after typing a partial filter — the common case — `isClearable` shows nothing and the only way back is backspacing. Separately, `noOptionsMessage={() => null}` removes the “No options” node, which is also what react-select’s live region announces, so a screen-reader user gets no feedback when a query matches nothing. **Fix:** render a clear control whenever `filter` is non-empty, and return a real string from `noOptionsMessage`.

**N8 — Nit — `menuPortalTarget={document.body}` puts the menu outside `.map-tab`.** [`FindspotFilterInput.tsx:60`](src/map/ui/FindspotFilterInput.tsx#L60), [`MapTab.sass:13-20`](src/map/ui/MapTab.sass#L13-L20). The portal is the right call — it keeps the menu above the MapLibre canvas’s stacking context. The consequence is that the `.map-tab__search .findspot-filter__*` rules can only ever style the control, never the menu: any future `.findspot-filter__menu` / `__option` rule added there will silently do nothing. Worth a top-level selector or a note when menu styling is first needed.

**N9 — Informational — `/tools/map` joins the generated sitemap; the committed archives will lag.** [`toolsConfig.tsx:16`](src/router/toolsConfig.tsx#L16), [`toolsRoutes.tsx:93`](src/router/toolsRoutes.tsx#L93). Routes are generated from `tabIds` with `{...(sitemap && sitemapDefaults)}`, so adding `'map'` is all that is needed — nicely done. It does mean `/tools/map` becomes a sitemap entry while `public/sitemap*.xml.gz` in the repo still predates it, until `update-sitemaps.yml` next runs. Nothing to change; just so the sitemap diff isn’t a surprise.

**N10 — Low — `provenances ?? []` in `MapTab` is unreachable, and is one of three uncovered branches.** [`MapTab.tsx:50`](src/map/ui/MapTab.tsx#L50). A full-suite `--coverage` run scoped to this PR’s changed production files puts `src/map` at **100% statements, functions and lines** with three uncovered branches: `MapTab.tsx:50` (`provenances ?? []`) — **not reachable**; `useFindspotMap.ts:154` (`error instanceof Error ? … : new Error(String(error))`) — reachable, untested; `FindspotFilterInput.tsx:50` (`option ? option.value : ''`) — reachable, untested, it is the clear-a-selected-suggestion path. The `MapTab.tsx` one can’t be covered because it can’t happen: `filterProvenances` returns `null` **iff** `provenances` is `null`, and `MapTab` returns the spinner while `filteredProvenances` is null, so past that early return `provenances` is always non-null and `?? []` never fires. Same shape as the old dead-`return` finding — a defensive guard that makes 100% unreachable while having no observable effect. **Fix:** drop the `?? []`, add a case constructing the map with a non-`Error` throw, and one clearing a selected suggestion. For the record, the fourth gap in the report — `toolsRoutes.entities.tsx:196`, the bibliography-editor route — is pre-existing; this PR only changed `import` to `import type` in that file.

PR #759, opened 2026-07-01
fix: pagination

Fabdulla1:
Add URL-backed pagination for Library search results. We use frontend-only query params such as `paginationIndex`. Keep pagination 0-based internally and display page numbers as 1-based. Clicking a pagination button should update URL and refreshing should preserve the active page.

### DIFF
--- a/src/fragmentarium/ui/search/FragmentariumSearchResult.test.tsx
+++ b/src/fragmentarium/ui/search/FragmentariumSearchResult.test.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { render, screen } from '@testing-library/react'
+import { render, screen, waitFor } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
 import { MemoryRouter, useLocation } from 'react-router-dom'
 import { SearchResult } from './FragmentariumSearchResult'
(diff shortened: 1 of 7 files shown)


khoidt (approved):
Hi! It works and it's clean. TypeScript compiles, lint is clean, and all 16 tests
pass with zero console noise. The one earlier review comment from the Codex bot — about page clicks mangling numeric search values like `number=000123` into `number=123` — has been properly fixed, with a regression test to keep it fixed.

**Pending findings, none of them blocking:**

- **Duplicated constant (low).** `paginationIndex` is spelled out as its own constant in two files —
  [fragmentariumRoutes.tsx:27](src/router/fragmentariumRoutes.tsx#L27) and
  [FragmentariumSearchResult.tsx:16](src/fragmentarium/ui/search/FragmentariumSearchResult.tsx#L16).
  One strips the key from the backend query, the other reads it, so if someone ever renames just
  one, pagination breaks quietly. Worth pulling into a single shared constant.

- **Two untested branches (low).** Both are defensive paths in new code that no test exercises: the
  "multiple `paginationIndex` params in the URL" case at
  [FragmentariumSearchResult.tsx:26](src/fragmentarium/ui/search/FragmentariumSearchResult.tsx#L26),
  and the duplicate-param cleanup at
  [PaginationItems.tsx:32](src/fragmentarium/ui/search/PaginationItems.tsx#L32). Two small tests would close this.

- **Cosmetic edge case (informational).** If a search returns fewer pages than the URL asks for, the view correctly snaps to the last page but the URL keeps the bigger number.

PR #762, opened 2026-07-03
Add Realia entries to sitemap via slugs

khoidt:
Extend the slug-driven sitemap to include Realia pages, mirroring existing structures for signs, dictionary, and bibliography. Introduce a new API endpoint to retrieve all Realia IDs and integrate it into the sitemap generation process.


Fabdulla1 (changes requested):
Overall, everything seems fine except 2 concerns I had. One concern is I think it would be good to test the behavior of `realiaService.listAllRealia()` returns an empty array. I think this would support against regression so sitemap generation continues to produce valid output without any Realia entry URLs. My second concern was adding a more targeted test to verify that `getEntityRoutes()` correctly thread `realiaSlugs` onto the `/tools/realia:id` route when sitemap generation is enabled. I think a focused unit test would make it easier to catch regressions if the route configuration is ever refactored in the future. Once these are added I think you can go ahead and merge :D

Fabdulla1 (approved):
Looks great! Thanks for adding the tests :)

PR #764, opened 2026-07-08
Final map

Fabdulla1:
Final map display with historical overlays and detailed data.


ejimsan (commented):
It's looking great! Here a few notes, will look at it again asap:

- In the search, Site name > Add suggestions, add collation (e.g. Assur will also find Aššur). Collation is implemented elsewhere in the application, take from there.
- Clicking on a site should zoom in enough to show the polygon (if it exists). At the moment, if the polygon is displayed and one clicks on the site, it actually zooms out.
- Throughout, speak of "tablets" instead of "fragments"
- Do not display the findspot ids, as they are arbitrary. Display instead a string of the values of the various fields, if necessary truncated

PR #765, opened 2026-07-09
Feature media architecture

Fabdulla1:
This PR adds an architecture-only frontend media foundation with domain types, DTOs, defensive mappers, repository and binary-loading contracts, and rollout documentation for future media integration. It preserves existing runtime image behavior while normalizing legacy hasPhoto / thumbnailPath compatibility and adding isolation tests to prevent premature production wiring.


khoidt (changes requested):
Nice work on this — the domain/DTO/mapper core is well designed, the legacy `hasPhoto`/`thumbnailPath` compatibility rules are carefully thought through, and the isolation test enforcing the no-runtime-change boundary is a good idea that most PRs of this kind skip. Lint and tsc are clean, and the Photo-tab wiring from the earlier commit was reverted properly.

Requesting changes on four hard gates:

- **250-line limit:** `mediaMapper.ts` is 377 lines and `mediaMapper.test.ts` is 415. Both need  splitting.
- **100% coverage:** `ImageViewer` is at 60% branch, `mediaMapper` at 91%. Every uncovered mapper line is a malformed-input guard — the mapper's whole purpose.
- **DRY:** `ImageViewer` re-implements `ButtonWithTooltip`, `handleDownload`, `handleOpenInNewTab` and the toolbar that already exist in `ImageButtonGroup`/`useImageActions`.
- **Green suite:** `FragmentList.test.tsx` fails intermittently. That one is pre-existing and unrelated to this PR — fixed separately in #771.

Two things beyond the gates that I'd like your take on. The isolation guard doesn't list
`common/ui/ImageViewer`, which is exactly why the Photo-tab wiring passed CI and had to be caught by a bot — the hole is still open. And `ImageViewer` breaks the SVG-new-tab and MIME-allowlist rules from `media-rollout-contract.md` in this same PR (it does `mimeType.split('/')[1]` and will `window.open` any blob, including an SVG, which runs script in our origin).

My suggestion: drop `ImageViewer` from this PR and land it with the deliberate runtime-integration work. That closes four of the findings at once and keeps this PR true to its own "architecture-only" contract.

Full review:

[TASK-765-review.md](https://github.com/user-attachments/files/30006273/TASK-765-review.md)

khoidt (changes requested):
Hi, good news first — everything mechanical is green and all four previous blockers are fixed. Dropping `ImageViewer` was the right call. However, I found some other issues that must be addressed before merging.

### Blockers

- **F1 — no way to cancel a request.** `MediaRepository.findByFragment` and `MediaBinaryLoader.fetch` take no `signal`, so nothing can cancel them. Today `withData`'s `if (fetchPromise?.cancel)` is silently false; after PR #774 (bluebird → `AbortController`) the signal has no parameter to travel through. Fix: add `signal?: AbortSignal` to both, keep the `Promise` return. Don't use `Bluebird` — it's being deleted.
- **F2 — `ThumbnailSize` defined twice**, in `domain/media.ts` and `FragmentService.ts:64`, with a test that watches them drift instead of preventing it. Pick one home, re-export, delete the test. 
- **F11 — `docs/` is the wrong place.** It doesn't exist on `master`, and the org's `generic-documentation` repo has an explicit table saying mapping changes go in `frontend-docs/02-data-and-api.md` and layering in `01-architecture.md`. The
  content is genuinely valuable — the SVG/download security rules and `legacyThumbnailPath` retirement criteria exist nowhere else — so **move** those, drop the PR-scoped bits (goals, non-goals, rollout steps 1-5) into the PR description, then delete `docs/`.

### Small stuff

- **F3** — `&& !normalizedPrimary` (`mediaSummaryMapper.ts:87`) can never be false; delete it.
- **F4** — the "barrel files" isolation test iterates an always-empty list. It can never fail.
- **F5** — the guard lists `common/ui/ImageViewer`, which no longer exists.
- **F6** — that guard's regex misses dynamic `import()` and `require()`.
- **F7** — `normalizeOptionalText` and `normalizeUrl` are just `normalizeNonEmptyString` wearing hats; `normalizeUrl` is public and promises validation it doesn't do.
- **F8** — `selectMediaById` sorts the whole list to find one unique id.
- **F9** — redundant `as MediaType` (`mediaSummaryMapper.ts:80`).
- **F10** — half the new files use alias imports, half use `./`. It's better stick to the former.

khoidt (changes requested):
Good round — F1 through F10 are genuinely fixed, and the split into `mediaSummaryMapper` / `mediaResourceMapper` / `mediaRepresentationMapper` / `mediaMapperValidation` reads much better than the single 377-line file. The AST-based isolation guard is a real improvement over the regex, the `signal?: AbortSignal` parameters landed exactly as discussed, and lint, the full test suite (418 suites, 4083 tests, zero console output) and qlty are all green.

Three things block merge, and none of them are in the mapper code.

- **F12** — this PR re-adds `docs/flaky-test-fragment-list-duplicate-key.md`, which `master` deleted in `8971a666`. The merge in `c1fb4540` resurrected it, so the diff now carries 129 lines of documentation about #771 (someone else's already-merged fix) and re-creates the very `docs/` directory F11 asked to remove.
- **F13** — F11 isn't resolved. The content moved from `docs/` to a *new local* `frontend-docs/` inside this repo, using the exact filenames `01-architecture.md` and `02-data-and-api.md` that already exist in the org's `generic-documentation` repo with broader, different content. I checked: the org copies still have nothing about media. So we now have two same-named docs in two repos, and the real ones never got the SVG/download rules.
- **F14** — `yarn tsc` is red on this branch. It's pre-existing (it fails on clean `master` too), but our own rules make it a hard gate, so it needs fixing before this merges.

Beyond that, three mapper behaviours look off to me — mostly around `legacyThumbnailPath` outliving the conditions that justify it, and the public `normalizeMediaSummary` dropping the "don't trust this" flag it computes internally. All three are cheap to fix and worth doing now, while there's still no runtime consumer to break. The rest are small.

### Details

#### F12 — Blocker — this PR resurrects a file `master` deleted

`docs/flaky-test-fragment-list-duplicate-key.md` (new in this PR's diff, 129 lines).

`master` added this file in `4ac457d8` (#771) and then deleted it in `8971a666`. This branch merged `master` at `c1fb4540` and the resolution kept the file, so `git diff $(git merge-base origin/master HEAD) HEAD` reports it as `A` (added). Two consequences: the PR now contains documentation about a *different* PR's already-merged test fix, and the `docs/` directory F11 asked to delete is still on the branch, holding exactly one unrelated file.

Fix: `git rm docs/flaky-test-fragment-list-duplicate-key.md` and remove the now-empty `docs/`. Worth double-checking the rest of the `c1fb4540` merge resolution for other files that `master` deleted and this branch brought back.

#### F13 — Blocker — F11 is still open; the docs went to the wrong place and shadow real filenames

`frontend-docs/01-architecture.md` and `frontend-docs/02-data-and-api.md` (both new, neither directory exists on `master`).

F11 asked for the durable content — the SVG/new-tab and download security rules, and the `legacyThumbnailPath` retirement criteria — to be moved into the org's `generic-documentation` repo, at `frontend-docs/01-architecture.md` and `frontend-docs/02-data-and-api.md`, with the PR-scoped bits (goals, non-goals, rollout steps) going into the PR description and `docs/` deleted. What happened instead is a rename: `docs/` became a new local `frontend-docs/` inside `ebl-frontend`, with the same two filenames the org repo already uses. I fetched `ElectronicBabylonianLiterature/generic-documentation` — `frontend-docs/01-architecture.md` and `frontend-docs/02-data-and-api.md` are there, and neither mentions media at all. So the security rules and retirement criteria still exist in exactly one place, and it's the wrong repo.

The filename collision makes it worse than a plain misplacement: the org's `01-architecture.md` documents the whole frontend architecture, while this branch's `01-architecture.md` is titled "Frontend Media Architecture" and covers only media. Anyone landing on either one has no way to know the other exists.

Also still outstanding from F11: `frontend-docs/01-architecture.md:16-27` keeps the PR-scoped `Goals` list that was meant to move into the PR description.

Fix: open a PR against `generic-documentation` folding the durable content into the existing `frontend-docs/01-architecture.md` (layering) and `frontend-docs/02-data-and-api.md` (DTOs, mapping, compatibility rules, security requirements), move `Goals` and the architecture-only boundary into this PR's description, then delete `frontend-docs/` and `docs/` from this branch.

#### F14 — Blocker (pre-existing) — `yarn tsc` fails

`src/fragmentarium/ui/text-annotation/realiaOptionLoader.ts:30` — `error TS18046: 'entries' is of type 'unknown'`.

Not introduced here — I reproduced it on a clean `origin/master` worktree, so it arrived with #767. But our instructions treat `yarn tsc` as a hard gate and require pre-existing failures to be fixed at their root cause rather than deferred, so it can't ride along.

Root cause: `RealiaService.search` returns a Bluebird `Promise<readonly RealiaEntry[]>`, and `toNativePromise<T>(promise: PromiseLike<T>)` (line 17) can't infer `T` from Bluebird's overloaded `then`, so `T` collapses to `unknown` and `entries` at line 30 is untyped.

Fix: bind the type argument explicitly — `toNativePromise<readonly RealiaEntry[]>(realiaService.search(query))` — or widen the parameter to accept Bluebird's promise type. Note that CI never catches this: no workflow runs `tsc`, which is why the PR is green.

#### F15 — Medium — a `count: 0` summary still hands the caller a thumbnail to render

`src/fragmentarium/infrastructure/mediaSummaryMapper.ts:148-159`.

When the new summary is valid and says there is no media at all, `hasPrimaryThumbnail` is false, so the legacy path is passed through anyway. Verified against the real module:

```
normalizeCompatibleMediaSummary({ mediaSummary: { count: 0, types: [] }, hasPhoto: true, thumbnailPath: '/legacy-thumbnail' })
→ { mediaSummary: { count: 0, types: [] }, legacyThumbnailPath: '/legacy-thumbnail' }
```

The caller is told "this fragment has zero media" and simultaneously handed a thumbnail URL. Rule 1 in `02-data-and-api.md` says a valid `mediaSummary` wins over legacy fields, and a `count: 0` summary is exactly that. No test covers this input.

Fix: return `legacyThumbnailPath: null` when the accepted summary has `count === 0`, and add the case to `mediaMapper.compatibility.test.ts`.

#### F16 — Medium — `legacyThumbnailPath` survives `hasPhoto: false`

`src/fragmentarium/infrastructure/mediaSummaryMapper.ts:126-134`, and the test that locks it in at `src/fragmentarium/infrastructure/mediaMapper.compatibility.test.ts:105-112`.

`normalizeLegacyMediaSummary` computes `legacyThumbnailPath` independently of `hasPhoto`, so a payload that explicitly declares "no photo" still yields a thumbnail path:

```
normalizeLegacyMediaSummary(false, '/legacy-thumbnail')
→ { mediaSummary: null, legacyThumbnailPath: '/legacy-thumbnail' }

normalizeCompatibleMediaSummary({ mediaSummary: { count: 1, types: ['PHOTO'] }, hasPhoto: false, thumbnailPath: '/legacy-thumbnail' })
→ { mediaSummary: { count: 1, types: ['PHOTO'] }, legacyThumbnailPath: '/legacy-thumbnail' }
```

The existing test is named "keeps thumbnail paths for legacy no-photo input", so this is deliberate — but I can't find the justification. In the legacy backend shape `thumbnailPath` only means anything when `hasPhoto` is true, and the whole point of `legacyThumbnailPath` is to feed the old thumbnail route. If there's a real payload where `hasPhoto: false` ships a usable `thumbnailPath`, could you write it into `02-data-and-api.md` as an explicit rule? Otherwise gate the path on `hasPhoto === true` and update the test.

#### F17 — Medium — the public entry point drops the "untrustworthy" flag it computes

`src/fragmentarium/infrastructure/mediaSummaryMapper.ts:120-124`.

`normalizeMediaSummaryWithDiagnostics` computes `hasCriticalError`, but `normalizeMediaSummary` — the function exported through the `mediaMapper` barrel — throws it away and returns only the summary. Result:

```
normalizeMediaSummary({ count: 5, types: ['NOPE'] })
→ { count: 5, types: [] }
```

Rule 5 in `02-data-and-api.md` calls "a positive count with no usable type or primary" not trustworthy enough to use, and `normalizeCompatibleMediaSummary` correctly falls back to legacy in that case. But a caller reaching for the plain `normalizeMediaSummary` gets the untrustworthy shell with no way to detect it — "5 media items, no types" would render as a broken gallery.

Fix: either return `null` for critical errors from `normalizeMediaSummary`, or export the diagnostics-carrying variant and keep the lossy one internal. `mediaMapper.summary.test.ts:128` currently pins the lossy behaviour, so it'd need updating with whichever choice you make.

#### F18 — Low — the isolation claim in the docs is no longer literally true

`frontend-docs/01-architecture.md:31-34` states the media modules "may be imported by tests and documentation examples only".

Since F2 was fixed, production `src/fragmentarium/application/fragmentServicePorts.ts:24` imports `fragmentarium/domain/media`. That's completely fine — it's `import type`, it's erased at build, and the guard deliberately skips type-only imports (`mediaArchitectureIsolationGuard.ts:56`) — but the doc now overstates the boundary, and the exception is the interesting part. One sentence saying type-only imports are permitted and why would keep the doc honest.

#### F19 — Low — the two contract tests exercise no production code

`src/fragmentarium/application/MediaRepository.test.ts` and `src/fragmentarium/application/MediaBinaryLoader.test.ts`.

Both modules are interfaces only, so there's nothing to execute at runtime. Each test declares a fake, calls it, and asserts it returned what it was literally written to return; `MediaBinaryLoader.test.ts:40-48` asserts `request.representation === 'display'` two lines after assigning it. The compile-time value is real — the interfaces do have to accept these shapes — but that's `tsc`'s job, and it's already covered by the fact that these files compile.

Not asking you to delete them without your agreement, but they'd carry their weight better as a single type-level assertion per module (or a `// @ts-expect-error` case proving a bad implementation is rejected) instead of five runtime tests that can't fail.

#### F20 — Low — one isolation test is fully subsumed by another

`src/fragmentarium/infrastructure/mediaArchitectureIsolation.test.ts:64-82` ("keeps every production barrel from re-exporting new media modules") filters production files to barrels and asserts no `reexport` references. The test at line 42 already asserts `findMediaArchitectureReferences(...)` equals `[]` for *every* production file and *every* reference kind, which includes those barrels and those re-exports. The second test can only fail if the first already has.

#### F21 — Low — the module inventory is asserted twice

`src/fragmentarium/infrastructure/mediaArchitectureIsolation.test.ts:34-38` asserts `mediaArchitectureModules` equals `findExpectedMediaArchitectureModules(sourceRoot)`, and `src/test-support/mediaArchitectureIsolationGuard.test.ts:92-108` hardcodes the same 10 module paths as the expected output of that same function. Adding a media module means editing the list in `mediaArchitectureIsolationGuard.ts` *and* the copy in the test. Keeping the derived-equals-declared assertion and dropping the hardcoded copy (or asserting shape/count instead) removes the duplicate.

#### F22 — Low — `MediaGalleryState` is dead

`src/fragmentarium/domain/mediaGallery.ts:3-5` — declared, never imported, never tested. Either it belongs with the runtime gallery PR that will actually hold state, or it should go now.

#### F23 — Low — `normalizeThumbnailSize` is the F7 pattern again

`src/fragmentarium/infrastructure/mediaRepresentationMapper.ts:92-96` is `isThumbnailSize(value) ? value : undefined` — the membership check it's credited with is entirely `isThumbnailSize`'s. Nothing outside `mediaMapper.representations.test.ts:127-130` calls it. `02-data-and-api.md:124-127` cites it as the example of a wrapper that earns its name, which makes it the one case the rule shouldn't be illustrated with. Callers can use the type guard directly.

#### F24 — Low — a barrel export that exists only for a test

`src/fragmentarium/infrastructure/mediaMapper.ts:17` re-exports `normalizeNonEmptyString`, and the only consumer is `mediaMapper.representations.test.ts:124-125`, which could import it from `mediaMapperValidation` directly. The barrel should describe the mapper's public surface, not the tests' import convenience.

Fabdulla1:
> Good round — F1 through F10 are genuinely fixed, and the split into `mediaSummaryMapper` / `mediaResourceMapper` / `mediaRepresentationMapper` / `mediaMapperValidation` reads much better than the single 377-line file. The AST-based isolation guard is a real improvement over the regex, the `signal?: AbortSignal` parameters landed exactly as discussed, and lint, the full test suite (418 suites, 4083 tests, zero console output) and qlty are all green.
> 
> Three things block merge, and none of them are in the mapper code.
> 
> * **F12** — this PR re-adds `docs/flaky-test-fragment-list-duplicate-key.md`, which `master` deleted in `8971a666`. The merge in `c1fb4540` resurrected it, so the diff now carries 129 lines of documentation about #771 (someone else's already-merged fix) and re-creates the very `docs/` directory F11 asked to remove.
> * **F13** — F11 isn't resolved. The content moved from `docs/` to a _new local_ `frontend-docs/` inside this repo, using the exact filenames `01-architecture.md` and `02-data-and-api.md` that already exist in the org's `generic-documentation` repo with broader, different content. I checked: the org copies still have nothing about media. So we now have two same-named docs in two repos, and the real ones never got the SVG/download rules.
> * **F14** — `yarn tsc` is red on this branch. It's pre-existing (it fails on clean `master` too), but our own rules make it a hard gate, so it needs fixing before this merges.
> 
> Beyond that, three mapper behaviours look off to me — mostly around `legacyThumbnailPath` outliving the conditions that justify it, and the public `normalizeMediaSummary` dropping the "don't trust this" flag it computes internally. All three are cheap to fix and worth doing now, while there's still no runtime consumer to break. The rest are small.
> 
> ### Details
> #### F12 — Blocker — this PR resurrects a file `master` deleted
> `docs/flaky-test-fragment-list-duplicate-key.md` (new in this PR's diff, 129 lines).
> 
> `master` added this file in `4ac457d8` (#771) and then deleted it in `8971a666`. This branch merged `master` at `c1fb4540` and the resolution kept the file, so `git diff $(git merge-base origin/master HEAD) HEAD` reports it as `A` (added). Two consequences: the PR now contains documentation about a _different_ PR's already-merged test fix, and the `docs/` directory F11 asked to delete is still on the branch, holding exactly one unrelated file.
> 
> Fix: `git rm docs/flaky-test-fragment-list-duplicate-key.md` and remove the now-empty `docs/`. Worth double-checking the rest of the `c1fb4540` merge resolution for other files that `master` deleted and this branch brought back.
> 
> #### F13 — Blocker — F11 is still open; the docs went to the wrong place and shadow real filenames
> `frontend-docs/01-architecture.md` and `frontend-docs/02-data-and-api.md` (both new, neither directory exists on `master`).
> 
> F11 asked for the durable content — the SVG/new-tab and download security rules, and the `legacyThumbnailPath` retirement criteria — to be moved into the org's `generic-documentation` repo, at `frontend-docs/01-architecture.md` and `frontend-docs/02-data-and-api.md`, with the PR-scoped bits (goals, non-goals, rollout steps) going into the PR description and `docs/` deleted. What happened instead is a rename: `docs/` became a new local `frontend-docs/` inside `ebl-frontend`, with the same two filenames the org repo already uses. I fetched `ElectronicBabylonianLiterature/generic-documentation` — `frontend-docs/01-architecture.md` and `frontend-docs/02-data-and-api.md` are there, and neither mentions media at all. So the security rules and retirement criteria still exist in exactly one place, and it's the wrong repo.
> 
> The filename collision makes it worse than a plain misplacement: the org's `01-architecture.md` documents the whole frontend architecture, while this branch's `01-architecture.md` is titled "Frontend Media Architecture" and covers only media. Anyone landing on either one has no way to know the other exists.
> 
> Also still outstanding from F11: `frontend-docs/01-architecture.md:16-27` keeps the PR-scoped `Goals` list that was meant to move into the PR description.
> 
> Fix: open a PR against `generic-documentation` folding the durable content into the existing `frontend-docs/01-architecture.md` (layering) and `frontend-docs/02-data-and-api.md` (DTOs, mapping, compatibility rules, security requirements), move `Goals` and the architecture-only boundary into this PR's description, then delete `frontend-docs/` and `docs/` from this branch.
> 
> #### F14 — Blocker (pre-existing) — `yarn tsc` fails
> `src/fragmentarium/ui/text-annotation/realiaOptionLoader.ts:30` — `error TS18046: 'entries' is of type 'unknown'`.
> 
> Not introduced here — I reproduced it on a clean `origin/master` worktree, so it arrived with #767. But our instructions treat `yarn tsc` as a hard gate and require pre-existing failures to be fixed at their root cause rather than deferred, so it can't ride along.
> 
> Root cause: `RealiaService.search` returns a Bluebird `Promise<readonly RealiaEntry[]>`, and `toNativePromise<T>(promise: PromiseLike<T>)` (line 17) can't infer `T` from Bluebird's overloaded `then`, so `T` collapses to `unknown` and `entries` at line 30 is untyped.
> 
> Fix: bind the type argument explicitly — `toNativePromise<readonly RealiaEntry[]>(realiaService.search(query))` — or widen the parameter to accept Bluebird's promise type. Note that CI never catches this: no workflow runs `tsc`, which is why the PR is green.
> 
> #### F15 — Medium — a `count: 0` summary still hands the caller a thumbnail to render
> `src/fragmentarium/infrastructure/mediaSummaryMapper.ts:148-159`.
> 
> When the new summary is valid and says there is no media at all, `hasPrimaryThumbnail` is false, so the legacy path is passed through anyway. Verified against the real module:
> 
> ```
> normalizeCompatibleMediaSummary({ mediaSummary: { count: 0, types: [] }, hasPhoto: true, thumbnailPath: '/legacy-thumbnail' })
> → { mediaSummary: { count: 0, types: [] }, legacyThumbnailPath: '/legacy-thumbnail' }
> ```
> 
> The caller is told "this fragment has zero media" and simultaneously handed a thumbnail URL. Rule 1 in `02-data-and-api.md` says a valid `mediaSummary` wins over legacy fields, and a `count: 0` summary is exactly that. No test covers this input.
> 
> Fix: return `legacyThumbnailPath: null` when the accepted summary has `count === 0`, and add the case to `mediaMapper.compatibility.test.ts`.
> 
> #### F16 — Medium — `legacyThumbnailPath` survives `hasPhoto: false`
> `src/fragmentarium/infrastructure/mediaSummaryMapper.ts:126-134`, and the test that locks it in at `src/fragmentarium/infrastructure/mediaMapper.compatibility.test.ts:105-112`.
> 
> `normalizeLegacyMediaSummary` computes `legacyThumbnailPath` independently of `hasPhoto`, so a payload that explicitly declares "no photo" still yields a thumbnail path:
> 
> ```
> normalizeLegacyMediaSummary(false, '/legacy-thumbnail')
> → { mediaSummary: null, legacyThumbnailPath: '/legacy-thumbnail' }
> 
> normalizeCompatibleMediaSummary({ mediaSummary: { count: 1, types: ['PHOTO'] }, hasPhoto: false, thumbnailPath: '/legacy-thumbnail' })
> → { mediaSummary: { count: 1, types: ['PHOTO'] }, legacyThumbnailPath: '/legacy-thumbnail' }
> ```
> 
> The existing test is named "keeps thumbnail paths for legacy no-photo input", so this is deliberate — but I can't find the justification. In the legacy backend shape `thumbnailPath` only means anything when `hasPhoto` is true, and the whole point of `legacyThumbnailPath` is to feed the old thumbnail route. If there's a real payload where `hasPhoto: false` ships a usable `thumbnailPath`, could you write it into `02-data-and-api.md` as an explicit rule? Otherwise gate the path on `hasPhoto === true` and update the test.
> 
> #### F17 — Medium — the public entry point drops the "untrustworthy" flag it computes
> `src/fragmentarium/infrastructure/mediaSummaryMapper.ts:120-124`.
> 
> `normalizeMediaSummaryWithDiagnostics` computes `hasCriticalError`, but `normalizeMediaSummary` — the function exported through the `mediaMapper` barrel — throws it away and returns only the summary. Result:
> 
> ```
> normalizeMediaSummary({ count: 5, types: ['NOPE'] })
> → { count: 5, types: [] }
> ```
> 
> Rule 5 in `02-data-and-api.md` calls "a positive count with no usable type or primary" not trustworthy enough to use, and `normalizeCompatibleMediaSummary` correctly falls back to legacy in that case. But a caller reaching for the plain `normalizeMediaSummary` gets the untrustworthy shell with no way to detect it — "5 media items, no types" would render as a broken gallery.
> 
> Fix: either return `null` for critical errors from `normalizeMediaSummary`, or export the diagnostics-carrying variant and keep the lossy one internal. `mediaMapper.summary.test.ts:128` currently pins the lossy behaviour, so it'd need updating with whichever choice you make.
> 
> #### F18 — Low — the isolation claim in the docs is no longer literally true
> `frontend-docs/01-architecture.md:31-34` states the media modules "may be imported by tests and documentation examples only".
> 
> Since F2 was fixed, production `src/fragmentarium/application/fragmentServicePorts.ts:24` imports `fragmentarium/domain/media`. That's completely fine — it's `import type`, it's erased at build, and the guard deliberately skips type-only imports (`mediaArchitectureIsolationGuard.ts:56`) — but the doc now overstates the boundary, and the exception is the interesting part. One sentence saying type-only imports are permitted and why would keep the doc honest.
> 
> #### F19 — Low — the two contract tests exercise no production code
> `src/fragmentarium/application/MediaRepository.test.ts` and `src/fragmentarium/application/MediaBinaryLoader.test.ts`.
> 
> Both modules are interfaces only, so there's nothing to execute at runtime. Each test declares a fake, calls it, and asserts it returned what it was literally written to return; `MediaBinaryLoader.test.ts:40-48` asserts `request.representation === 'display'` two lines after assigning it. The compile-time value is real — the interfaces do have to accept these shapes — but that's `tsc`'s job, and it's already covered by the fact that these files compile.
> 
> Not asking you to delete them without your agreement, but they'd carry their weight better as a single type-level assertion per module (or a `// @ts-expect-error` case proving a bad implementation is rejected) instead of five runtime tests that can't fail.
> 
> #### F20 — Low — one isolation test is fully subsumed by another
> `src/fragmentarium/infrastructure/mediaArchitectureIsolation.test.ts:64-82` ("keeps every production barrel from re-exporting new media modules") filters production files to barrels and asserts no `reexport` references. The test at line 42 already asserts `findMediaArchitectureReferences(...)` equals `[]` for _every_ production file and _every_ reference kind, which includes those barrels and those re-exports. The second test can only fail if the first already has.
> 
> #### F21 — Low — the module inventory is asserted twice
> `src/fragmentarium/infrastructure/mediaArchitectureIsolation.test.ts:34-38` asserts `mediaArchitectureModules` equals `findExpectedMediaArchitectureModules(sourceRoot)`, and `src/test-support/mediaArchitectureIsolationGuard.test.ts:92-108` hardcodes the same 10 module paths as the expected output of that same function. Adding a media module means editing the list in `mediaArchitectureIsolationGuard.ts` _and_ the copy in the test. Keeping the derived-equals-declared assertion and dropping the hardcoded copy (or asserting shape/count instead) removes the duplicate.
> 
> #### F22 — Low — `MediaGalleryState` is dead
> `src/fragmentarium/domain/mediaGallery.ts:3-5` — declared, never imported, never tested. Either it belongs with the runtime gallery PR that will actually hold state, or it should go now.
> 
> #### F23 — Low — `normalizeThumbnailSize` is the F7 pattern again
> `src/fragmentarium/infrastructure/mediaRepresentationMapper.ts:92-96` is `isThumbnailSize(value) ? value : undefined` — the membership check it's credited with is entirely `isThumbnailSize`'s. Nothing outside `mediaMapper.representations.test.ts:127-130` calls it. `02-data-and-api.md:124-127` cites it as the example of a wrapper that earns its name, which makes it the one case the rule shouldn't be illustrated with. Callers can use the type guard directly.
> 
> #### F24 — Low — a barrel export that exists only for a test
> `src/fragmentarium/infrastructure/mediaMapper.ts:17` re-exports `normalizeNonEmptyString`, and the only consumer is `mediaMapper.representations.test.ts:124-125`, which could import it from `mediaMapperValidation` directly. The barrel should describe the mapper's public surface, not the tests' import convenience.

Yes! I'd like to mention that the docs will be removed, I just had them here as guiding/progress tracking for myself. I will be removing them!

khoidt (changes requested):
Fourth pass, and this is the best round yet. Every single finding from round 3 that lives in the code is genuinely fixed — F12, F13, F17, F18, F20, F21, F22, F23 and F24 all check out, and F19 is most of the way there. `docs/` and `frontend-docs/` are both gone from the branch, the `mediaMapper` barrel now describes only its real public surface, `normalizeMediaSummary` returns `null` instead of quietly handing back an untrustworthy shell, and the isolation guard's tests stopped duplicating the module inventory. The split into four focused mapper modules has aged well — every file is comfortably under the 250-line limit, the largest being a test at 226 lines.

The mechanical picture is mostly green: `yarn lint` is clean, qlty is clean on all three axes (checks, structure smells, duplication), diff coverage is 100%, GitHub's checks are green, and the suite produced zero console output across every run. There are no `.md` files in the diff, and — I checked this specifically, commit by commit as well as against the merge base — **no dev container, Docker, CI workflow, or dependency configuration is touched anywhere in this PR**. Nothing there needs careful scrutiny because there is nothing there.

Two blockers, though. `yarn tsc` is still red, same as last round. And the full suite is not green: `src/dictionary/ui/editor/DerivedFromInput.test.tsx` failed on a run where it had passed forty minutes earlier in the same session. It is a genuine flake and not this PR's fault — the same species as the `FragmentList` flake from round 1 — but a green suite is a hard gate, and I traced it to a root cause that is a one-line fix in a shared test fixture.

Beyond that there is one behaviour I would like changed before it sets, and it is the sharper version of F15/F16. When the backend sends `mediaSummary: { count: 0 }` alongside `hasPhoto: true`, the mapper reports zero media — the photo simply vanishes. Curiously, if the backend omits `mediaSummary` entirely, the same fragment correctly comes back as one PHOTO. So a *present-but-empty* summary is treated as more authoritative than an explicit `hasPhoto: true`, which is exactly backwards for a backfill window. That is now pinned by a test, so it is deliberate — but the rationale F16 asked for was going to live in `02-data-and-api.md`, and deleting the docs (rightly!) took its home with it. Right now the only statement of this contract is a test name.

The rest are small: a test whose name promises more than it asserts, a percent-encoded traversal that slips past the URL trust boundary, and a couple of redundancy nits.

### Details

#### P1 — Blocker (pre-existing) — `DerivedFromInput.test.tsx` is flaky; unseeded random fixtures can emit duplicate values

`src/dictionary/ui/editor/DerivedFromInput.test.tsx:31-36` ("Displays all notes"), root cause in `src/test-support/word-fixtures.ts:67-69`.

Observed in this review: the file **passed** in one run and **failed** in another, in the same session, against the same commit, with no code change in between. The failure is `TestingLibraryElementError: Found multiple elements with the display value: ih`, with two `<input>` elements (`TextInput-17`, `TextInput-18`) both carrying `value="ih"`.

Root cause is `wordArray`, which every word fixture builds on:

```ts
function wordArray(chance = defaultChance) {
  return [chance.word(), chance.word()]
}
```

`defaultChance` is an unseeded `new Chance()`, and the two draws are independent with no uniqueness constraint, so the array can legitimately contain the same word twice. `DerivedFromInput` renders one `<input>` per note, and the test's `screen.getByDisplayValue(note)` at line 34 throws the moment two notes match, because `getBy*` requires exactly one hit. Measured over 200,000 trials, `chance.word()` collides with itself about **0.016%** of the time (32/200,000) — roughly one run in 6,000 — because `chance.word()` draws 2-character words about 17% of the time from a small effective pool (~129,000 distinct words across 200,000 draws). Rare enough to look green for months, frequent enough to fail a CI run eventually. `formFactory`, `derivedFactory`, `logogramFactory` and `wordFactory` all share `wordArray`, so any test doing `getByDisplayValue` over one of these arrays carries the same latent flake.

Fix at the root, not the symptom — make the array's elements distinct rather than switching the assertion to `getAllByDisplayValue`, which would only hide it:

```ts
function wordArray(chance = defaultChance) {
  return chance.unique(chance.word, 2)
}
```

This keeps the data random, fixes every consumer of `wordArray` at once, and needs no change to the test itself. Seeding `defaultChance` deterministically would be a defensible alternative, but it is a much larger behavioural change across all fixture files (ten modules construct their own `new Chance()`), so I would not do it as part of this.

**I have not applied this fix.** It touches `src/test-support/word-fixtures.ts`, which has nothing to do with media architecture, and landing it here would put an unrelated change in this PR's diff — exactly the problem F12 was raised about. It belongs on its own branch off `master`, the way the `FragmentList` flake was handled in #771. Flagging it as a blocker because our instructions treat a green suite as a hard gate and require pre-existing failures to be fixed at their root cause rather than deferred; where that fix lands is a separate question from whether it lands.

#### F14 — Blocker (pre-existing) — `yarn tsc` fails

`src/fragmentarium/ui/text-annotation/realiaOptionLoader.ts:30` — `error TS18046: 'entries' is of type 'unknown'.`

Still exactly as reported in round 3, and still not introduced by this branch: `git show origin/master:src/fragmentarium/ui/text-annotation/realiaOptionLoader.ts` is byte-identical to the branch copy, so a clean `master` fails the same way. No CI workflow runs `tsc`, which is why every check on this PR is green while the gate is red.

Root cause is unchanged. `RealiaService.search` returns a Bluebird promise, and `toNativePromise<T>(promise: PromiseLike<T>)` at line 17 cannot infer `T` through Bluebird's overloaded `then`, so `T` collapses to `unknown` and `entries` at line 30 is untyped.

Fix: bind the type argument explicitly — `toNativePromise<readonly RealiaEntry[]>(realiaService.search(query))` — or widen the parameter to Bluebird's promise type. Our own instructions require pre-existing failures to be fixed at their root cause in the same task rather than deferred, so this cannot ride along a fourth time.

#### R1 — Medium — a present-but-empty `mediaSummary` silently overrides `hasPhoto: true`

`src/fragmentarium/infrastructure/mediaSummaryMapper.ts:150-162`, pinned by `src/fragmentarium/infrastructure/mediaMapper.compatibility.test.ts:201-225`. This supersedes F15 and F16, which are both still open.

`normalizeCompatibleMediaSummary` treats any new summary that is non-null and free of critical errors as authoritative. `{ count: 0, types: [] }` qualifies — `hasCriticalError` is `false` for it — so the legacy branch is never consulted, and the `hasPhoto: true` signal is dropped on the floor. Verified by executing the real compiled module:

```
normalizeCompatibleMediaSummary({ mediaSummary: { count: 0, types: [] }, hasPhoto: true,  thumbnailPath: '/fragments/K.1/thumbnail/small' })
→ { mediaSummary: { count: 0, types: [] }, legacyThumbnailPath: '/fragments/K.1/thumbnail/small' }

normalizeCompatibleMediaSummary({ mediaSummary: { count: 0, types: [] }, hasPhoto: false, thumbnailPath: '/fragments/K.1/thumbnail/small' })
→ { mediaSummary: { count: 0, types: [] }, legacyThumbnailPath: '/fragments/K.1/thumbnail/small' }

normalizeCompatibleMediaSummary({                                        hasPhoto: true,  thumbnailPath: '/fragments/K.1/thumbnail/small' })
→ { mediaSummary: { count: 1, types: ['PHOTO'] }, legacyThumbnailPath: '/fragments/K.1/thumbnail/small' }
```

Two things fall out of this. First, the first two results are identical: once an empty new summary is present, `hasPhoto` has no effect whatsoever, which is precisely the F16 concern about `legacyThumbnailPath` being computed independently of `hasPhoto`, now visible on the compatibility path rather than the legacy one. Second, and worse, cases one and three describe the *same fragment* — one photo, one thumbnail route — yet the fragment that got an empty summary reports `count: 0`. During the backfill window this is the common case, and a consumer that gates on `mediaSummary.count > 0` renders nothing for a fragment that demonstrably has a photo.

I accept that this is deliberate — the test names say so ("keeps the legacy thumbnail route while the new media projection is empty before backfill", "treats thumbnailPath as an unconditional route hint, not gated by hasPhoto"). My problem is that the contract those names imply — *`count` may be `0` while the fragment actually has media, so always check `legacyThumbnailPath` too* — is genuinely surprising, and a test name is not where a cross-cutting rule for every future consumer should live. F16 asked for it in `02-data-and-api.md`; the F13 fix correctly deleted that file, so the rationale now has nowhere to be.

Preferred fix: when the new summary is valid but empty and `hasPhoto === true`, fall back to the legacy summary, so the un-backfilled fragment reports the photo it has. If you would rather keep current behaviour, then it needs to be stated somewhere durable — the `generic-documentation` PR that F13 asks for is the natural home, and `NormalizedMediaSummaryCompatibility` is the natural anchor. Either way, `mediaMapper.compatibility.test.ts:201-225` needs updating to match the decision. Worth settling now, while there is still no runtime consumer to break.

#### R2 — Low — a test whose name promises distinctness it never asserts

`src/fragmentarium/application/MediaBinaryLoader.test.ts:40-51`.

The case is titled "addresses a distinct route for the %s representation" and runs across all five representations, but each invocation builds its own `requested` array and asserts only `expect(requested[0]).toContain('/fragments/K.1/media/media-id/')` — the prefix every representation shares. Nothing is ever compared *across* representations, so the case has no way to observe distinctness even in principle: it would pass unchanged if `original`, `display` and all three thumbnail sizes collapsed onto a single URL, which is the exact regression the name implies it guards.

The exact-route assertions already exist, and correctly, in `src/fragmentarium/infrastructure/mediaUrls.test.ts:71-100`. So either assert distinctness here (`expect(new Set(urls).size).toBe(representations.length)` after looping) or drop the case and let `mediaUrls.test.ts` own route shape. The rest of this file is a real improvement on round 3 — routing the fake through `fragmentMediaBinaryUrl` means it now exercises production code, which is what F19 was asking for.

#### R3 — Low — percent-encoded traversal passes the URL trust boundary

`src/fragmentarium/infrastructure/mediaMapperValidation.ts:17-28`.

`normalizeRelativeMediaUrl` rejects traversal by splitting on `/` and looking for a literal `..` segment, so it only sees traversal that is already decoded. Verified against the compiled module:

```
'/fragments/../admin'      → undefined   (rejected, as intended)
'/fragments/%2e%2e/admin'  → '/fragments/%2e%2e/admin'   (accepted)
'/fragments/%2E%2E/admin'  → '/fragments/%2E%2E/admin'   (accepted)
'/fragments/.%2e/admin'    → '/fragments/.%2e/admin'     (accepted)
'/fragments/K.1/a\nb'      → '/fragments/K.1/a\nb'       (accepted)
'/'                        → '/'                          (accepted)
```

Impact is limited — these paths are same-origin and backend-supplied, so the worst case is addressing a different endpoint on our own API rather than crossing an origin boundary — which is why this is Low and not a blocker. But this module has a dedicated test file called "relative media url trust boundary" whose stated job includes rejecting `../admin`, and a validator that catches the plain spelling while waving through the encoded one gives more assurance than it earns. Percent-decoding each segment before the `..` comparison closes it without breaking the legitimately-encoded case (`/fragments/K.1%2F2/media/media-id/file` must keep passing). The stray newline and bare-`/` acceptances are the same shape of gap and worth folding into the same fix.

#### R4 — Low — `MediaRepository.test.ts` still asserts only what its own fake was written to return (F19 residue)

`src/fragmentarium/application/MediaRepository.test.ts:47-61`.

`MediaBinaryLoader.test.ts` answered F19 by routing through `fragmentMediaBinaryUrl`, so it now exercises real code. Its sibling did not: the fixture at lines 4-33 hardcodes `/fragments/K.1/media/photo-id/file`, and line 55-60 asserts that the fake returned those same strings. Nothing under `src/` runs. The compile-time value is real — the interface does have to accept this shape — but that is `tsc`'s job and is already proven by the file compiling.

Same suggestion as round 3, and still not something I would change without your agreement: either build the fixture URLs with `fragmentMediaOriginalUrl(...)` so the test is anchored to production routing the way its sibling now is, or reduce it to a type-level assertion.

#### R5 — Low — the isolation test walks and reads the entire source tree twice per run

`src/fragmentarium/infrastructure/mediaArchitectureIsolation.test.ts:43-53` and `56-61`.

The first case calls `listSourceFiles(sourceRoot)` and reads every production file; the second calls `listSourceFiles(sourceRoot)` again and reads *every* source file, tests included, to regex for barrels. That is two full walks of ~1100 files and roughly two full reads of the tree on every suite run. Hoisting the file list and the sources into a `beforeAll`-scoped constant shared by both cases keeps the semantics identical for a fraction of the I/O.

#### R6 — Low — the on-disk existence loop is subsumed by the assertion below it

`src/fragmentarium/infrastructure/mediaArchitectureIsolation.test.ts:23-31`.

`findExpectedMediaArchitectureModules` only ever returns modules it found on disk, so the very next case — `expect(mediaArchitectureModules).toEqual(findExpectedMediaArchitectureModules(sourceRoot))` at lines 34-38 — already proves every listed module exists. The `existsSync` loop cannot fail unless that equality has failed first. The `expect(mediaArchitectureModules.length).toBeGreaterThan(0)` on line 21 is *not* redundant and should stay: it is what stops both cases going vacuous if the list and the tree ever empty out together. Dropping just the loop is the whole change.

#### R7 — Info — nothing in production reaches any of this yet

Eleven modules and 2835 lines are, by design, reachable from production only through the `ThumbnailSize` type re-export in `src/fragmentarium/application/fragmentServicePorts.ts:24` — which is erased at build time. That is the PR's own stated contract and the isolation guard enforces it well, so this is not a finding. Flagging it only so the merge decision is made with it in mind: the 94.2% coverage number is entirely test-driven, and none of these mapping rules have met a real backend payload yet. R1 is worth settling precisely because that first real consumer is the moment these choices stop being free.

khoidt (approved):
This one is done. Every finding from round 4 is genuinely fixed — I re-ran each one against the compiled code rather than reading the diff — and both blockers turned out to be fixed on master and inherited here, so `yarn tsc` and the full suite are green for the first time in five rounds. 450 suites, 4422 tests, zero failures, zero console output, 100% diff coverage, qlty and CodeQL clean. No `.md` files, and **no dev container, Docker, CI or dependency file is touched anywhere in this PR**.

Approving. One thing to settle, not worth another round: unknown media types are handled three different ways, and two of them lose data silently — so the day a third `MediaType` is added, `count` and `types` will disagree and the gallery will render fewer items than the summary claims. Nothing in production can reach this code yet, which is exactly why it is cheap to decide now and expensive after the integration PR.

## Findings

### R8 — Medium — an unknown `MediaType` is handled four different ways, three of them silent

`src/fragmentarium/infrastructure/mediaSummaryMapper.ts:31-34` and `75-85`; `src/fragmentarium/infrastructure/mediaResourceMapper.ts:93-102`.

`normalizeMediaTypes` filters with `isMediaType` and keeps what survives. `MediaTypes` is `['PHOTO', 'COPY']`, so any third value the backend introduces is dropped. What happens next depends entirely on the shape it arrived in. Verified by executing the modules:

```text
normalizeMediaSummary({ count: 1, types: ['LINE_DRAWING'] })
→ null                                          (treated as corrupt)

normalizeMediaSummary({ count: 2, types: ['PHOTO', 'LINE_DRAWING'] })
→ { count: 2, types: ['PHOTO'] }                (silently trimmed)

normalizeMediaSummary({ count: 1, types: ['PHOTO'], primary: { id: 'x', type: 'LINE_DRAWING' } })
→ { count: 1, types: ['PHOTO'] }                (the whole primary, thumbnail included, vanishes)

normalizeFragmentMediaResponse({ media: [ {id:'a', type:'PHOTO', ...}, {id:'b', type:'LINE_DRAWING', ...} ] }).media.map(r => r.id)
→ ['a']                                         (the resource is dropped from the gallery)
```

And through the compatibility path, only the first of those sets the diagnostic:

```text
normalizeCompatibleMediaSummary({ mediaSummary: { count: 1, types: ['LINE_DRAWING'] }, hasPhoto: true, ... })
→ { mediaSummary: { count: 1, types: ['PHOTO'] }, ..., newSummaryIsMalformed: true }
```

So the same unknown token is a corruption signal in one shape and a no-op in the other three, and `newSummaryIsMalformed` — the flag added in `14fc016e` precisely to say "do not trust this" — stays `false` for every case that quietly loses data.

The concrete consequences on the day a third type ships:

- `count` and `types` disagree. A consumer building type-filter chips from `types` shows "Photo" only, and one of the two items is unreachable through any filter while the header still says 2.
- `primary` disappears for a fragment whose primary is the new type, so the summary thumbnail falls back or blanks, with nothing indicating why.
- The gallery renders fewer items than the summary counted — `normalizeFragmentMediaResponse` drops the resource, `normalizeMediaSummary` keeps the count.

This is the same species as R1 from round 4, and the same argument applies: it is a contract decision that is free to make now and expensive once a consumer exists. It is also *only* reachable through a backend change, so it will surface as "the new media type doesn't show up" long after this code ships, which is the worst time to diagnose it.

The current tests pin the unknown-only case well (`mediaMapper.summary.test.ts:124-137`, `mediaMapper.compatibility.test.ts:157-169`) but no test covers the mixed case, the unknown-primary case, or the dropped-resource case — which is how the inconsistency stayed invisible.

Preferred fix — make "I dropped something I did not recognise" one signal rather than four behaviours. `normalizeMediaSummaryWithDiagnostics` already computes everything needed:

```ts
const droppedTypes = (types as readonly unknown[]).length - normalizedTypes.length
const droppedPrimary = primary !== undefined && primary !== null && normalizedPrimary === undefined
```

Surface either as part of the existing `newSummaryIsMalformed` flag, or as a sibling `hasUnrecognizedMedia`, so a consumer can decide between rendering a partial gallery and telling the user their client is out of date. The narrow alternative is to treat any unknown type as critical everywhere, matching today's unknown-only behaviour — more conservative, and defensible while there is no consumer, but it means one unrecognised item hides a fragment's entire media set, so I would not choose it.

Either way the decision wants a test per shape, so the next person reads it as deliberate. I have not applied a change.

### R9 — Minor — residual traversal encodings still pass the URL trust boundary

`src/fragmentarium/infrastructure/mediaMapperValidation.ts:42-54`.

Every case I raised in round 4 is fixed — the segment-wise `decodeURIComponent` plus the control-character check closed all of them, and the legitimate `%2F` case still passes. What remains is narrower:

```text
'/fragments/%2e%2e%2fadmin'  → '/fragments/%2e%2e%2fadmin'   (one segment, decodes to '../admin')
'/fragments/....//admin'     → '/fragments/....//admin'
'/a/%252e%252e/b'            → '/a/%252e%252e/b'             (double-encoded)
```

The first is the one worth closing. `url.split('/')` splits on *literal* slashes, so `%2e%2e%2fadmin` stays one segment; `decodeURIComponent` then yields `../admin`, which is not equal to `'..'`, so the check misses it. Whether that resolves as traversal depends on the server — a compliant one keeps `%2F` literal, but proxies and some configurations normalise it first.

Impact stays Low for the same reasons as round 4: these paths are same-origin and backend-supplied, so the worst case is addressing a different endpoint on our own API. But this module has a test file called "relative media url trust boundary", and a validator that catches `%2e%2e/` while waving through `%2e%2e%2f` is claiming a little more than it delivers.

Fix: compare against the traversal segment after splitting the *decoded* segment on slashes too, rather than comparing the decoded segment whole:

```ts
const decodedSegments = segments.flatMap((segment) => {
  const decoded = decodeUrlSegment(segment)
  return decoded === undefined ? [undefined] : decoded.split('/')
})
```

That closes `%2e%2e%2f` and `....//` alike without affecting `/fragments/K.1%2F2/media/media-id/file`, which decodes to `K.1/2` — two segments, neither of them `..`. Double-encoding is a genuine non-issue unless something decodes twice; I would leave it.

### R10 — Minor — MIME type matching is case-sensitive

`src/fragmentarium/infrastructure/mediaRepresentationMapper.ts:42-44`, via `isRasterMediaMimeType` / `isOriginalMediaMimeType` in `domain/media.ts:90-109`.

```text
normalizeRasterRepresentation({ url: '/a/b', mimeType: 'image/jpeg' })  → { url: '/a/b', mimeType: 'image/jpeg' }
normalizeRasterRepresentation({ url: '/a/b', mimeType: 'IMAGE/JPEG' })  → undefined
normalizeRasterRepresentation({ url: '/a/b', mimeType: 'Image/Jpeg' })  → undefined
```

RFC 2045 §5.1 makes the type and subtype case-insensitive, so `IMAGE/JPEG` is a valid `image/jpeg`. Here it silently drops the representation — and for `original`, that drops the entire resource via `normalizeMediaRepresentations` returning `undefined`.

To be fair about severity: the direction is fail-safe. Nothing dangerous gets through; a legitimate image just disappears. And our own backend presumably emits lowercase, so this is latent rather than live. But it is one `.toLowerCase()` in the two predicates, it costs nothing, and the failure mode — an image that silently does not render, with no diagnostic — is unpleasant to debug. Note that `' image/jpeg '` already works, since `normalizeNonEmptyString` trims, so the module is half-way to being lenient about formatting already.

Parameters (`'image/jpeg; charset=x'`) are also rejected. I would leave that alone: stripping parameters is a real parse, and rejecting is the right default.

### R11 — Info — the isolation guard flags fully-type-only named imports

`src/test-support/mediaArchitectureIsolationGuard.ts:54`.

```ts
if (ts.isImportDeclaration(node) && !node.importClause?.isTypeOnly) {
```

`isTypeOnly` is true only for `import type { X } from '...'`. The inline form where *every* specifier is type-only — `import { type MediaResource } from 'fragmentarium/domain/media'` — has `importClause.isTypeOnly === false`, so it is reported as a violation even though TypeScript erases it completely.

The guard's own tests are consistent with the current behaviour and correct about the mixed case: `import { type MediaResource, isMediaType } from '...'` genuinely does import a value and is rightly flagged (`mediaArchitectureIsolation.test.ts:70-73`). It is the all-type-only-specifiers case that is a false positive, and it is untested.

Fail-safe direction, and nothing in the tree writes imports that way today, so this is a note rather than a finding. It will matter at integration time, when production files legitimately start importing these types and someone reaches for the inline form. Closing it means checking that every named binding carries `isTypeOnly`, roughly:

```ts
const namedBindings = node.importClause?.namedBindings
const allSpecifiersAreTypeOnly =
  namedBindings !== undefined &&
  ts.isNamedImports(namedBindings) &&
  namedBindings.elements.length > 0 &&
  namedBindings.elements.every((element) => element.isTypeOnly) &&
  node.importClause?.name === undefined
```

### R12 — Info — the guard's discovery rule needs a retirement plan

`src/test-support/mediaArchitectureIsolationGuard.ts:19-20` and `126-138`.

`isMediaArchitectureFile` classifies any file under `fragmentarium/` whose basename matches `/^media/i` as an architecture module, and `findExpectedMediaArchitectureModules` cross-checks the hardcoded inventory against that scan. That is a good design for the current phase — it is what makes the inventory test non-vacuous, and it is why R6's `existsSync` loop was safe to delete.

The consequence to plan for: the moment someone adds a genuine production component named `MediaGallery.tsx` or `MediaTab.tsx` under `fragmentarium/ui/`, the inventory test fails until it is added to the list at lines 5-17, and adding it there then forbids production from importing it — which is the opposite of what an integration PR wants. Nothing to change now; worth a line in the integration PR's description so it is a decision rather than a surprise. The guard should probably be scoped to an explicit path list, or deleted outright, when the runtime wiring lands.

### R13 — Info (pre-existing) — CI action deprecations

`.github/workflows/` — two `warning` annotations on the `Analyze (javascript)` check run, identical on this PR and on #816.

- "CodeQL Action v3 will be deprecated in December 2026" — `github/codeql-action/{init,autobuild,analyze}@v3` need `v4`.
- "Node.js 20 is deprecated. The following actions target Node.js 20 but are being forced to run on Node.js 24: `actions/checkout@v4`, `github/codeql-action/*@v3`" — the same bump plus `actions/checkout@v5`.

Repo-wide, and this PR touches no workflow file. Recorded because it appears in this PR's checks and will eventually break a run on `master`. Belongs on its own branch.

### R14 — Info — still zero production reachability

Carried over from round 4's R7, restated because it is the frame for the merge decision rather than a defect.

2932 lines across eleven modules are reachable from production only through the `ThumbnailSize` type re-export at `src/fragmentarium/application/fragmentServicePorts.ts:24`, which is `import type` + `export type` and therefore erased at build. The isolation guard enforces that, and I re-verified the guard is not vacuous: it walks 1400-odd real source files and separately proves it can find barrel re-exports, so the scan is doing work.

The honest reading of the coverage number: every media module with executable code is at 100%, but that is entirely test-driven. None of these mapping rules has met a real backend payload. R8 matters precisely because the first real consumer is the moment these choices stop being free.

PR #766, opened 2026-07-10
fix: concurrency bug

Fabdulla1:
fixing a quick refresh/search result bug

### DIFF
--- a/src/common/utils/ConcurrencyLimiter.test.ts
+++ b/src/common/utils/ConcurrencyLimiter.test.ts
@@ -117,7 +117,7 @@ describe('ConcurrencyLimiter', () => {
     let activeWork = 0
     let maximumActiveWork = 0
 
-    const activeOperations = activeCompletions.map((completion, index) =>
+    const activeOperations = activeCompletions.map((completion) =>
       limiter.run(() => {
         activeWork += 1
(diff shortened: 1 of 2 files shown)


khoidt (changes requested):
Hi, the fix is correct — I reproduced the original deadlock on `master` (a cancelled queued request leaks its slot forever until the limiter saturates and all fetches stop) and confirmed it's gone here. Lint, tsc, and the full suite are green.

Two things before merge:

1. Coverage is 68.75% branch, below our 100% gate — and more tests won't fix it: lines 38 and 65 are unreachable. Removing them also lets the `active` flag and the `while` loop in `releaseSlot()` go, since `acquire()` always succeeds there.
2. No test cancels an in-flight operation — only queued ones. That path works solely because Bluebird's `.finally()` runs on cancellation, which is load-bearing and untested; rewriting it as `.then()/.catch()` would silently reintroduce this exact leak.

Minor: unused `index` at `ConcurrencyLimiter.test.ts:120`, and `:160`'s `toBeLessThanOrEqual(2)` passes even if nothing ran — should be `toBe(2)`.

Full review attached:

[TASK-766-review.md](https://github.com/user-attachments/files/30003199/TASK-766-review.md)

PR #767, opened 2026-07-10
Add realia annotation layer to named entity annotation

khoidt:
Realia annotations run in parallel to the existing entity layer (PN, DN, ...) and reuse the same UI components for editing and display. A realia span is `{id, span, realiaId}`; `realiaId` is the source of truth and all embedded display data (`tier`, `name`, `layer`) is derived client-side and stripped before saving by the single helper `omitDerivedSpanFields`.

## The feature

- Extract the entity | realia span union, guards and id generation into `annotationSpan.ts`. Entity and realia ids are numbered independently.
- Generalise the annotation reducer and tiering over both layers, so a realia span and an entity span covering the same tokens stack on separate tiers.
- Add `RealiaSelect`, an async picker over `RealiaService.search` keyed on `realiaId`, wired into `SpanAnnotator` and `SpanEditor`. Its loader (`realiaOptionLoader.ts`) debounces, discards stale responses by request id, routes rejections into an empty option list rather than an unhandled rejection, and cancels on unmount.
- Split `Markable` (245 lines) into `SpanIndicator` and `InlineEditor`, with the presentation both the editor and the read-only view need factored into the `useSpanIndicator` hook.
- Resolve realia ids through the deployed `/realia/by-id/` endpoint, so `/tools/realia/{realiaId}` works for both lemmas and realia ids. `RealiaService.find` caches each entry under both keys.
- Link annotated tokens to their realia entry in the read-only transliteration view, honouring `RouterLinkModeContext` so the Word/PDF exports render nothing.
- Load the editor from the fragment response alone: `createFragmentAnnotationSpans` derives both span lists from `fragment.namedEntities`, `fragment.realia` and the word tokens. `GET /fragments/{number}/named-entities` is no longer called, and `fetchNamedEntityAnnotations` is removed from `FragmentService`, `ApiFragmentRepository` and the repository port.
- Report persistence and refresh separately. `updateNamedEntityAnnotations` resolves `{fragment, refreshError}`, so a refresh that fails *after* a successful save advances the saved baseline and shows a warning, instead of reporting the save itself as failed and leaving the editor dirty.
- Cover `clearSelection`, which previously had no tests on either browser branch.

## Blocked on ebl-api #740

**This PR must not merge before [ebl-api #740](https://github.com/ElectronicBabylonianLiterature/ebl-api/pull/740) is merged and deployed.**

The client now posts `{ namedEntities, realia }` to `POST /fragments/{number}/named-entities`, where it previously posted `{ annotations: [...] }`. On the currently deployed API, `named_entities.py`'s `on_post` still reads `req.media["annotations"]`, so **saving any named-entity annotation raises `KeyError` and returns a 500** — including annotations that have nothing to do with realia.

Reading degrades gracefully in the meantime: the deployed `FragmentSchema` already dumps `named_entities` under `data_key="namedEntities"`, so existing annotations still load; `realia` and `realiaInfo` are simply absent and fall back to empty.

#740 adds `RealiaAnnotationSpanSchema`, the `{ namedEntities, realia }` payload keys, and `realiaInfo` on every fragment-returning route. The client field names (`id`, `type`, `span` for entity spans; `id`, `realiaId`, `span` for realia spans; `realiaId`, `lemma`, `type` for `realiaInfo`) match that schema exactly, with no client-side aliases.

## Changes outside the realia feature

Beyond the feature itself, this PR carries a few unrelated changes, so they do not come as a surprise in review:

- **`src/signs/ui/display/`** — `yarn tsc` was already failing on `master` with two `TS18046` errors in `SignImages.tsx`, because inferring `R` from the Bluebird returned by `getClusterVariants` collapsed to `unknown`. `SignImages.tsx` is split into `SignImageFigures.tsx`, `signClusterAnnotations.ts` and `signImageGrouping.ts`; the fix itself is `task: (item: T) => PromiseLike<R>` in `signImageGrouping.ts` plus explicit type arguments at the call site. Unrelated to realia; fixed under the repo's "fix pre-existing issues" rule.
- **`src/test-support/utils.ts`, `src/test-support/waitForSpinnerToBeRemoved.ts`, `src/signs/ui/display/SignInformation.test.tsx`** — test infrastructure only: type the `testDelegation` target without `lodash.isFunction`, give the spinner wait an explicit timeout, and await the rendered links in the sign-information test so it cannot race.

## Line counts

**Every `.ts`/`.tsx` file in the changed set is at or under the 250-line ceiling.** Six files were already over it on `master` and were split as part of this PR rather than left as follow-up work:

| File | before | now | extracted into |
| --- | ---: | ---: | --- |
| `FragmentService.ts` | 729 | 246 | `fragmentServiceBase`, `fragmentCache`, `scopedCache`, `fragmentCacheKeys`, `fragmentProvenance`, `fragmentReferences` |
| `FragmentRepository.ts` | 541 | 234 | `fragmentFactories`, `fragmentRepositoryUpdates`, `fragmentRepositoryAttestations` |
| `SignImages.tsx` | 368 | 199 | `signClusterAnnotations`, `signImageGrouping`, `SignImageFigures` |
| `FragmentService.test.ts` | 1921 | 211 | 13 sibling suites + 3 test-support modules |
| `FragmentRepository.test.ts` | 950 | 171 | 6 sibling suites |
| `test-fragment.ts` | 560 | 126 | `test-fragment-dto`, `test-fragment-lines`, `test-fragment-more-lines` |

`fragment.ts` (214) replaces `Fragment`'s 34-parameter positional constructor with a single `FragmentProps` argument.


Fabdulla1 (approved):

Fabdulla1 (changes requested):
Everything looks great, just small change that needs to be addressed. The control in `src/fragmentarium/ui/display/FragmentDisplaySettings.tsx` shows and hides both named-entity and Realia annotations but the title and accessible label still say "Toggle named entities". Maybe it might be worth it to change to Toggle annotations or "Toggle named entitiesa nd Realia" so it can accurately describe behavior for screen readers. Other than that, it might be worth it to add a test that is a workflow test. Maybe a focused test confirming that both arrays are submitted together and that derived fields such as `tier`, `name`, and `layer` are stripped.

Fabdulla1 (changes requested):
Looking good so far! I mainly have test concerns:

`src/injectedApp.testSupport.tsx` introduces eslint-disable-next-line @typescript-eslint/no-explicit-any alongside `jest.MockedClass<any>`. Do you think it's possible to replace it with a typed generic or a narrower mocked-constructor type without disabling the rule? 

Also noticed that `RealiaSelect.tsx` calls the search promise with `.then(callback)` but doesn't handle rejection. The tests don't cover search failure or the race where query A starts, query B resolves first, and query A resolves afterward. I think it would be good to add explicit error handling and behavior level tests that prove a rejected search doesn't create an unhandled promise rejection, loading state is cleared appropriately, a stale result can't replace results from a newer query, unmounting during a request doesn't update state, and existing selection remains intact after a search failure. 

I also noticed that the annotation editor updates its local saved baseline only after the complete parent save/refresh promise resolves. If persistance succeeds but the following refresh fails then I think the UI may report a failure and remain dirty even though the annotation was saved. Maybe we could separate the outcomes and add focused tests that cover where persistence succeeds and refresh succeeds, persistence succeeds and refresh fails, persistence fails, retrying after a refresh-only failure doesn't duplicate or overwrite data, a stale refresh response can't replace newer local edits. 

`useAnnotationContext` initializes its reducer from `words` and `initial` only during the initial mount. I think if refreshed annotation data or a different fragment is supplied without remounting the component then the reducer can retain stale local state. Maybe we could demonstrate that the component always remounts when its source fragment changes? Or we could add synchronization keyed to the fragment/annotation identity with tests or refreshed props? 

Last but not least, the current implementation appears to keep annotation UI out of Word and PDF output through provider placement, but tests don't assert that the export output excludes annotation indicators, links, roles, and focusable elements. If possible, please add focused negative assertions for both Word and PDF rendering so a future provider-placemenet change won't silently expose interactive annotation UI in exports.

Thanks :D

Fabdulla1 (approved):
It's looking fantastic! Before merging, just make sure to remove the comments in `fragmentQueryMapping.ts` and `RealiaEntry.ts`. Also, the read-only Realia indicators currently use a `<span role="link" tabIndex={0}>`. Keyboard activation works but then native browser link behavior such as middle-click, context-menu actions, copying the link address, and opening a new tab are unavailable. Maybe the read-only view renders a real anchor when the target URL is already know while keeping the current interactive span behavior only where editing requires it? 
Last but not least I noticed the `waitForSpinnerToBeRemoved` now applies a 3000 ms timeout globally to every test using the shared helper. Before merging could you score the longer timeout to the specific test or workflow that requires it or explain why the existing default is insufficient across the entire suite? Thanks :D

PR #770, opened 2026-07-14
Fix n calls

Fabdulla1:
This PR reduces frontend API fan-out for Fragmentarium and research-project searches by introducing bounded server-side pagination and render-ready summary responses. It also limits dictionary fragment queries while preserving accurate result counts, prevents inactive photo and folio tabs from loading media, and serves summary thumbnails through their API URLs.
For Latest Additions, this PR adds client support for prefetched fragments, but the production /fragments/latest endpoint does not yet provide them. Consequently, the front page and /library will continue making per-item fragment requests until the corresponding backend change is implemented and deployed.
Finally, this PR removes the empty library search page from the sitemap and adds canonical URLs for fragment pages and folio variants.


khoidt (changes requested):
Nice one — this is a genuinely useful direction! Three things I'd want sorted before this goes in, though. First, the Latest Additions N+1 fix doesn't actually fire: summary fragments always come back with `record: []`, so the `hasLatestTransliterationRecord` guard sends every card straight back into `fragmentService.find`. The test that's supposed to protect it hand-builds a summary item with a record, which the repository can never produce — so it passes while the optimisation is inert. Second, switching search to `count: 'page'` quietly drops the line-match totals: I checked against the live API and `count=exact` still returns `matchCountTotal: 90` for `transliteration=me lik`, while `count=page` returns `null`, so the results header degrades from "Found 90 lines in 77 documents" to "Found 50 documents" — and that 50 is just the page size now. Third, the lemma-examples block on word pages: `RenderFragmentLines` never slices to `linesToShow`, and the summary preview now carries *every* matching line — `IM.48816` comes back with 16 preview lines, so the page renders all 16 and then says "And 13 more" underneath.

The rest is smaller stuff: image tabs now refetch on every switch, the lemma-examples header count can disagree with what's rendered, and a couple of files went over the 250-line ceiling. All detailed below with repro steps. Happy to re-review as soon as the blockers are addressed. 🙂

### Details

Numbers below match the `Findings` section further down, where each item has full evidence and repro steps.

**Blockers**

- **1 — Latest Additions N+1 removal never engages** (High) — `createQuerySummaryFragment` always builds the fragment with `record: []`, so `hasLatestTransliterationRecord` is false for every real summary item and all five preview cards fall through to `fragmentService.find`. The test guarding this builds a summary item *with* a record, a shape `FragmentRepository` cannot produce. Live `/fragments/latest` returns plain items today, so the preview issues 5 `find` calls either way.
- **2 — `count: 'page'` drops the line-match totals** (High) — live, `count=exact` returns `matchCountTotal: 90` for `transliteration=me lik` while `count=page` returns `null`, so the header degrades from "Found 90 lines in 77 documents" to "Found 50 documents". Worse, that document count is now `data.items.length`, i.e. the page size, constant across pages. Line queries also never get the "Showing results X–Y" range.
- **5 — Lemma examples render every matching line** (High) — `RenderFragmentLines` never slices to `linesToShow`, and the summary preview carries all matching lines. `IM.48816` returns 16 preview lines, so the word page renders 16 and appends "And 13 more". `totalLines` also comes from different sources in the two branches (`matchCount` vs. `lineIndexes.length`).

**Hard gates**

- **6 — Coverage below 100% on affected code** (Medium) — qlty diff coverage 96.0%. Four measured gaps in new code: `LatestTransliterations.tsx:37` (summary thumbnail `onError`), `FragmentariumSearchResult.tsx:18` (non-integer/negative `offset` fallback), `FragmentariumSearchResult.tsx:95` (singular `line` branch), `PaginationItems.tsx:26` (disabled-click guard). `encodeFragmentNumber`'s `catch` is covered only incidentally, via `chance.string()`.
- **13 — 250-line ceiling exceeded** (Medium) — `LatestTransliterations.tsx` 230 → 311 and `Images.test.tsx` 217 → 310 were pushed over **by this PR**; `LatestTransliterations.test.tsx` (556), `FragmentariumSearch.test.tsx` (547), `FragmentView.test.tsx` (322) and `FragmentRepository.test.ts` (1017) were already over and grew further.
- **4 — DRY: `10` duplicated in `FragmentLemmaLines.tsx`** (Medium) — the cap is written at line 212 (`limit: 10`) and line 138 (`_.take(items, 10)`), and the header counts the *unclipped* list, so 12 returned items read "Showing 12 Library document examples" while only 10 render.
- **7 — DRY: `https://www.ebl.lmu.de` hardcoded a fifth time** (Low) — `createFragmentCanonicalUrl` duplicates a literal that already exists in `sitemap.tsx` (`DOMAIN`), `chapter.ts`, `DictionaryLineGroup.tsx`, `DictionaryLineVariant.tsx`, `WordExport.tsx` and `PdfExport.tsx`.

**Everything else**

- **3 — Image tabs refetch on every switch** (Medium) — `TabPane` now unmounts inactive panes, and `findFolio`/`findPhoto`/`folioPager` are uncached, so folio A → B → A re-requests A's image and pager.
- **8 — Two different `getPageIndex` functions** (Low) — `pagination.ts:51` takes a search string, `FragmentariumSearchResult.tsx:17` takes a `FragmentQuery`.
- **9 — `encodeFragmentNumber` double-decodes** (Low) — `'A%41B'` becomes `'AAB'`, linking to a different fragment.
- **10 — Out-of-range `paginationIndex` no longer normalised** (Low) — `?paginationIndex=9999` strands the user on an empty page instead of clamping to the last one.
- **11 — Redundant "; more results are available"** (Low) — shown exactly when the Next button is already enabled.
- **12 — Pagination loses page jumping and result size** (Low) — Previous / "Page N" / Next only; `count: 'exact'` was added to `FragmentQuery` but is never used.
- **14 — `parseFragmentSearchQuery` called twice per render** (Info) — `fragmentariumRoutes.tsx:119-123` parses the same search string twice.
- **15 — Global mutation in `FragmentView.test.tsx`** (Info) — `HelmetProvider.canUseDOM = false` at module scope changes behaviour for every test in the file.
- **16 — Deleted tests: justified, no action needed** (Info) — the removed `PaginationItems` / chunking cases target code paths this PR deletes; `getActivePageFromSearch` has no remaining references.

**Green:** `yarn lint`, `yarn tsc` and the full suite (339 suites, 3481 passed, 2 pre-existing skips) all pass with **zero console output**. All 8 CI checks and all 3 qlty statuses are green.

khoidt (changes requested):
Big improvement since the last round — most of the previous findings are genuinely fixed, and the test files that were bloated are now less than half their old size. Nice work. Three things still block, though.

The first two are the same two blockers as last time, and unfortunately both are still live. The Latest Additions N+1 removal still doesn't fire: I checked `/fragments/latest` against the live API and it returns plain items (just `museumNumber`, `matchingLines`, `matchCount`) — no `fragment`, no  `thumbnailPath` — so every card falls straight through to `fragmentService.find` exactly as before. The new tests hand-build summary items that the endpoint can't produce, so they pass while the optimisation does nothing. And the line-count fix has swapped one bug for another: switching line queries to `count: 'exact'` does restore `matchCountTotal`, but `count=exact` returns `hasNextPage: null` on the live API, so the **Next** button is disabled on page 1 of every transliteration and lemma search — I confirmed there really are 27 more documents behind that disabled button for `transliteration=me lik`.

Third, the suite is no longer console-clean: `FragmentariumSearch.test.tsx` emits an `act(...)` warning from `withData`'s `setData` in both full runs I did.

The rest is smaller: diff coverage dropped to 82.7%, a few tests were deleted whose code paths are still very much alive (the "Did you mean" suggestion now has no test at all), the CDLI default-tab change can select a tab that has no nav item, and one file is still over the 250-line ceiling. All detailed below with repro steps. Happy to re-review as soon as the blockers are sorted. 🙂

### Details

Numbering matches the `Findings` section further down, where each item repeats the evidence in full.

**Blockers**

- **1 — Line searches can never reach page 2** (High, new — introduced by the fix for the previous review's finding 2) — `createPagedFragmentQuery` sends `count: 'exact'` for transliteration/lemma queries, and the live API returns `hasNextPage: null` for `count=exact`. `SearchResult` renders `hasNextPage={data.hasNextPage === true}`, which is therefore `false`, so `PaginationItems` disables **Next** on the first page of every line search. Live evidence: `GET /api/fragments/query?transliteration=me%20lik&limit=50&offset=0&count=exact` → `{matchCountTotal: 90, isMatchCountTotalExact: true, hasNextPage: null, items: 50}`; the same query at `offset=50` returns 27 more items, so page 2 demonstrably exists; the same query with `count=page` returns `hasNextPage: true`. The new "Go to page" box bypasses `hasNextPage`, so a user who guesses that page 2 exists can still get there, but nothing in the UI indicates it does.
- **2 — Latest Additions N+1 removal is still inert, and the `/library` list still hydrates per card** (High, previous finding 1 not fixed) — live `GET /api/fragments/latest` returns 49 items whose only keys are `matchCount`, `matchingLines`, `museumNumber`. On the front page (`preview={true}` from `Introduction.tsx:278`) `CompactFragmentCard` sees no `queryItem.fragment` and falls through to `HydratedCompactFragmentCard`, issuing five `fragmentService.find` calls. On `/library` (`Fragmentarium.tsx:63`) `FragmentLines` is given `includeLatestRecord={true}`, and `hasRenderReadyFragment` additionally requires a non-historical `Transliteration` record — which `createQuerySummaryFragment` can never supply because it builds the fragment with `record: []` (`FragmentRepository.ts:263`) — so that surface hydrates per item too. Neither surface reduces any calls today. Separately, if the backend ever does start returning summaries, the preview path will silently lose the transliteration date, because `CompactFragmentCardContent` derives `recordDate` from `fragment.uniqueRecord` with no equivalent guard — precisely the regression the Codex bot raised in the (now resolved, now outdated) thread on `LatestTransliterations.tsx:184`. The two new tests, `preview renders repository-shaped summaries without detail hydration` and `preview caps repository-shaped summaries before rendering cards`, both construct `QueryItem`s carrying a full `fragment`, a shape `/fragments/latest` does not return, so they pass while the code under test never runs in production.
- **3 — The test suite is no longer console-clean** (High, hard gate) — `src/fragmentarium/ui/search/FragmentariumSearch.test.tsx` emits `console.error: Warning: An update to ComponentWithData inside a test was not wrapped in act(...)`, originating at `src/http/withData.tsx:50` (`setData`). The React component stack is `FragmentariumSearch` → `Tabs` → `TabPane` → `ComponentWithData`, i.e. the Corpus tab's `textService.query` promise settling after the test's assertions have finished. Reproduced in both full-suite runs; it does **not** reproduce when the file is run on its own, which is consistent with a floating promise rather than a deterministic ordering bug. Per the review guidelines a PR may not be approved while any console-noise finding remains, and suppressing the warning is not an acceptable fix — the test needs to await or otherwise settle the corpus-tab request.

**Hard gates**

- **4 — Diff coverage is 82.7%, against a 100%-on-affected-code gate** (Medium) — qlty reports `qlty coverage diff | success | 82.7% (75% threshold)`, down from 96.0% at the previous review; it passes qlty's own 75% bar but not the project gate. Measured local gaps in code this PR adds or changes: `LatestTransliterationCard.tsx` (new file) 63.6% — the whole `HydratedCompactFragmentCard` getter, which is the only path that actually executes against the live API, plus the entire `LatestAdditionThumbnail` blob component and the `recordDate` / `periodAbbr` / projects branches; `SummaryThumbnail.tsx` (new file) 60.0% — the `!thumbnailPath || isBroken` early return and the `onError` handler; `pagination.ts` — `updatePageSizeSearchParam` is entirely uncovered and `getPageIndexForOffset`'s non-integer fallback is uncovered; `PaginationItems.tsx` — the `disabled` click guard and the whole "Results per page" `onChange`; `withData.tsx` — the new `onRetry` callback, so although `retry: true` was added to `FolioImage`, `FolioPager` and `FragmentPhoto`, the Retry button is never exercised; `FragmentLink.tsx` — the surrogate-pair and lone-low-surrogate branches of `toWellFormed`; `FragmentariumSearchResultComponents.tsx` — `hasLatestTransliterationRecord`, i.e. the very guard added to address the previous finding 1; `FragmentariumSearchResult.tsx` — the `Found 0 documents` branch.
- **5 — Tests were deleted whose code paths this PR does not remove** (Medium, hard gate: existing tests must not be removed without explicit confirmation) — `FragmentariumSearch.test.tsx` went from 547 to 181 lines and lost, among others: `Shows suggestion when entering wrong number format`, although `showNumberSuggestion` is still there at `FragmentariumSearchResult.tsx:150` and a repo-wide grep finds no remaining test asserting "Did you mean" anywhere — this is also the direct cause of the uncovered `Found 0 documents` branch; `Displays corpus results when clicking corpus tab` and `Updates URL anchor when clicking tab`, although corpus-tab switching and the `#corpus` hash are untouched by this PR and now have no test. Two snapshot files were also deleted (`FragmentariumSearch.test.tsx.snap`, 1143 lines, and `LatestTransliterations.test.tsx.snap`, 11155 lines). By contrast the `PaginationItems` numbered-page and chunking tests are legitimately obsolete, since that UI no longer exists — no objection to those.
- **6 — One file still exceeds the 250-line ceiling** (Medium) — `src/fragmentarium/ui/search/FragmentariumSearchResult.test.tsx` is 252 lines. Everything else changed by this PR is now under the limit, which is a real improvement: `LatestTransliterations.test.tsx` 556 → 109, `FragmentariumSearch.test.tsx` 547 → 181, and `LatestTransliterations.tsx` 311 → 115 after `LatestTransliterationCard.tsx` was extracted.

**Everything else**

- **7 — The default image tab can select a tab that has no nav item** (Medium, new) — `TabController.defaultKey` now includes `this.fragment.getExternalNumber('cdliNumber') && CDLI` (`Images.tsx:46`), but the CDLI nav item is rendered on `fragment.cdliImages && fragment.cdliImages.length > 0` (`Images.tsx:157`) while the CDLI pane is rendered on `getExternalNumber('cdliNumber')` (`Images.tsx:191`). For a fragment that has a `cdliNumber`, no `cdliImages`, no photo and at least one folio, the default key becomes `cdli`: no nav item is highlighted, every folio tab renders unselected, and an empty CDLI pane is shown instead of the first folio. Before this PR the first folio opened. Either align all three conditions on the same predicate or drop CDLI from `defaultKey`.
- **8 — The canonical domain is still hardcoded in five modules** (Low, previous finding 7 only partially fixed) — `router/domain.ts` was added and is now used by `sitemap.tsx` and `FragmentLink.tsx`, but the literal `https://www.ebl.lmu.de` still appears in `corpus/domain/chapter.ts:165`, `dictionary/ui/search/DictionaryLineGroup.tsx:34`, `dictionary/ui/search/DictionaryLineVariant.tsx:37`, `fragmentarium/ui/fragment/PdfExport.tsx:747` and `fragmentarium/ui/fragment/WordExport.tsx:174-175`. (The occurrences under `about/ui/newsletter/*.md.ts` are prose and should stay as they are.)
- **9 — The empty-page redirect makes the "No results on this page" UI unreachable** (Low) — the new effect in `FragmentariumSearchResult.tsx` calls `history.replace` with `paginationIndex=0` whenever a page past the first returns no items and `hasNextPage !== true`. In the real app that means the `'No results on this page'` string and the enabled Previous button — both asserted by `keeps a usable Previous control for an empty directly linked page`, which only passes because the test renders `SearchResult` directly with a fixed `fragmentQuery` and so never refetches — can never actually be seen; the user is bounced back to page 1 and loses their position. The previous behaviour clamped to the *last* page, which preserved it. Either clamp instead of resetting, or remove the now-dead branch and its test.
- **10 — Line searches no longer show a document count at all** (Low) — the header went from `Found 90 lines in 77 documents` to `Found 90 matching lines. Showing documents 1-50`. The previous review was right that `77` had silently become the page size, but the fix drops the document total rather than restoring a correct one, so that information is simply gone.
- **11 — `updatePageSizeSearchParam` re-encodes the whole query string** (Info) — it builds its result with `URLSearchParams.toString()`, whereas the neighbouring `updatePaginationSearchParam` was deliberately written to preserve the original encoding. Changing the page size therefore rewrites `:` to `%3A` and spaces to `+` across every parameter. I verified this is harmless in practice — `query-string.parse` round-trips both forms to identical objects — but the two helpers in the same module now disagree on encoding style for no stated reason.
- **12 — `TextAnnotation.test.tsx` is close to its timeout budget** (Info, not attributable to this PR) — `Named Entity Annotation › calls updateNamedEntityAnnotations on save` failed with `Exceeded timeout of 5000 ms for a test` in run 2, which I had inadvertently overlapped with a running dev server. It passed in run 1 and passes 3/3 in isolation, and the file is not touched by this PR, so this is an artifact of the CPU contention I introduced rather than a defect here. Mentioned only because the suite takes ~40s and the margin on that test is evidently thin enough to matter on a loaded CI runner.

khoidt (changes requested):
Really nice round — almost everything from the last two reviews is genuinely fixed this time, and I could verify most of it against the live API rather than just reading code. The line-search pagination fix is clever and it works: overfetching `limit + 1` and deriving `hasNextPage` yourself neatly sidesteps the backend returning `null` for `count=exact`, and I confirmed page 2 is now reachable for `transliteration=me lik`. The lemma-examples clipping, the image-tab caching, the CDLI tab alignment, the canonical-domain module, the 250-line splits — all sorted. The suite is console-clean again (349 suites, 3517 tests, zero output), and diff coverage is back to 100%. The new mapper-driven tests for Latest Additions are exactly the right answer to my "the tests build a shape the repository can't produce" complaint. Thank you for that.

Two things block, though, and one of them is new.

The new one is the bigger worry: `ResultPages` no longer chunks, so every `SearchResult` call site that *doesn't* go through `createPagedFragmentQuery` now renders its entire result set in one go. That's the four research-project search pages, which this PR doesn't touch and no test covers. `/projects/CAIC/search` with an empty form asks the API for `project=CAIC` with no `limit`, gets **15,752 items** back, and tries to render 15,752 fragment cards — where `master` rendered 50 with numbered pagination. Next is permanently disabled there too, because the API returns `hasNextPage: null` without a `count` param. I confirmed this by rendering `SearchResult` with exactly the props those pages pass (`fragmentQuery={{ project: 'CAIC' }}`, no `resultPageSize`) against a 137-item result with `hasNextPage: null`: all **137** rendered, **Next disabled**, header `Showing documents 1-137`. I'd treat that as a hard blocker.

The second is the one carried over: the Latest Additions fan-out still doesn't actually go away, because `GET /fragments/latest` on production still returns bare items — just `museumNumber`, `matchingLines`, `matchCount`, no `fragment`, no `thumbnailPath`, and no top-level `fragments` array. The client side is correct and well tested now; it's simply waiting on a backend change that hasn't shipped. So the front page still makes 5 `find` calls and `/library` still makes ~50. Worth confirming that the API change is actually planned before this merges, since it's the PR's headline claim.

Everything else is small — a lost "Found N documents" on plain searches, a couple of redundant mechanisms, and a pre-existing `yarn tsc` failure that isn't yours. Details below, with repro steps. Happy to re-review as soon as the two blockers are sorted. 🙂

### Details

Numbering matches the `Findings` section further down, where each item repeats the evidence in full.

**Blockers**

- **1 — Research-project search pages lost pagination and now render every result at once** (High, new — introduced by this PR). `ResultPages` used to do `const pages = _.chunk(fragments, RESULTS_PER_PAGE)` and render `pages[displayActive]`; it now renders `fragments.map(...)` directly, and `SearchResult` only slices when `isLineQuery` is true (`visibleItems = isLineQuery ? data.items.slice(0, visiblePageSize) : data.items`, `FragmentariumSearchResult.tsx:94-96`). `/library/search` is fine because `createPagedFragmentQuery` always injects `limit`/`offset`/`count`. But `src/research-projects/subpages/ResearchProjectSearch.tsx:45` and `src/research-projects/subpages/Home.tsx:49` pass a bare `fragmentQuery` with no `limit`, no `offset`, no `count` and no `resultPageSize`, and `src/router/researchProjectRoutes.tsx:101-102` builds it as `{...parse(location.search), project}`. None of those three files is touched by this PR, and none of them has a test. Live evidence: `GET https://www.ebl.lmu.de/api/fragments/query?project=CAIC` returns **15,752 items** with `hasNextPage: null`. So `/projects/CAIC/search` renders 15,752 `FragmentLines` subtrees where `master` rendered 50. On top of that, `effectiveHasNextPage = data.hasNextPage ?? (isLineQuery && …)` evaluates to `false` for `null`, so **Next** is permanently disabled; the new "Go to page" box pushes `paginationIndex=N` into the URL, which the route forwards straight to the API as an unrecognised parameter (the API accepts and ignores it) while `offset` is never set, so the page never changes. The "Results per page" select does reach the API as `limit`, but that truncates the whole result set rather than paginating it. Four projects are affected: CAIC, aluGeneva, AMPS, RECC.
- **2 — The Latest Additions fan-out removal is still inert against the production API** (High, previous round-2 finding 2, partially fixed). The *test-realism* half of that finding is genuinely fixed: `LatestTransliterations.mapper.test.tsx` now drives the real `FragmentRepository.queryLatest()` mapper instead of hand-building `QueryItem`s, and `preview hydrates mapper-produced summary placeholders for latest record data` (line 147) honestly asserts that summary items still hydrate. What remains is a pure backend dependency. Live `GET https://www.ebl.lmu.de/api/fragments/latest` returns top-level keys `["matchCountTotal","items"]` and 50 items whose only keys are `matchingLines`, `museumNumber`, `matchCount`. `createLatestQueryResult` (`FragmentRepository.ts:295-320`) can only attach a fragment from (a) a summary item — `isQuerySummaryItemDto` requires `description`, `script` and `hasPhoto`, none of which are present — or (b) a top-level `fragments` array, which the API does not send. So `queryItem.fragment` is `undefined` for every item, `hasRenderReadyFragment` is false, and both surfaces hydrate per card: the front page issues 5 `fragmentService.find` calls (`Introduction.tsx` → `preview={true}` → `LatestTransliterations.tsx:46-52` → `CompactFragmentCard` → `HydratedCompactFragmentCard`) and `/library` issues one per item (`LatestTransliterations.tsx:72-82` with `includeLatestRecord={true}`). Note also that even if the backend started sending *summary* items, this path still could not short-circuit, because `includeLatestRecord: true` requires a non-historical `Transliteration` record and `createQuerySummaryFragment` hardcodes `record: []` (`FragmentRepository.ts:242`, `:263`) — only the `fragments` prefetch array can satisfy it. Per the project's own rule that the backend schema is the source of truth, adding a `fragments` field the API does not emit needs backend confirmation before this lands.

**Hard gates**

- **3 — `yarn tsc` is red** (Medium, **pre-existing — not introduced by this PR**). Two errors: `src/signs/ui/display/SignImages.tsx(297,41)` and `(306,42)`, both `TS18046: 'result.value' is of type 'unknown'`. I reproduced them at the PR's merge base `4ac457d8`, and the file is not in this PR's diff, so they are inherited. For completeness, `origin/master` is *also* red, with a different error — `src/fragmentarium/ui/text-annotation/realiaOptionLoader.ts(30,9): 'entries' is of type 'unknown'` — verified at both `aea1dca3` and the current tip `e8e51cc3`, so rebasing swaps one failure for another rather than clearing the gate. Root cause on `SignImages.tsx`: `runWithConcurrencyLimit<T, R>` declares `task: (item: T) => Promise<R>` (line 229-233), but the call site at line 290 passes `signService.getClusterVariants(...)`, which returns a Bluebird promise; inference against the native `Promise<R>` shape fails and `R` collapses to `unknown`. This is not cosmetic — `craco start` surfaces both as full-screen error-overlay entries while webpack itself reports "Compiled successfully!".

**Everything else**

- **4 — Plain (non-line) searches lost the "Found N documents" total** (Medium, new). `resultInfo = isLineQuery ? lineResultInfo : documentRange` (`FragmentariumSearchResult.tsx:128`), and `documentRange` only ever produces `Showing documents X-Y`. `master` rendered `Found {fragmentCount} document(s)`. The code that would restore it already exists — `isCompleteFirstPage` and `pageDocumentCount` (lines 106-109) compute exactly `Found N documents` for the case where page 1 is the complete result set — but it is wired only into `lineResultInfo`, i.e. only for line queries whose response lacks `matchCountTotal`. Since line queries now send `count: 'exact'` and the live API always returns `matchCountTotal` for that, that branch is effectively unreachable in production. The existing test `disables Previous on page zero and disables Next on the last page` (`FragmentariumSearchResult.test.tsx:49-58`) locks the regression in by asserting `Showing documents 1-12` for `hasNextPage: false` on page 0.
- **5 — `PdfExport.tsx` is still 776 lines** (Low, pre-existing). It was 775 on `master` and this PR nudged it to 776 while changing `getHyperLink`. It is the only changed file over the 250-line ceiling, and every other file this PR touches is now under it — including `LatestTransliterations.tsx` 230 → 115, `FragmentariumSearch.test.tsx` 546 → 83 and `FragmentView.test.tsx` 291 → 111, which is a real improvement. Flagging only because the ceiling is a hard gate and the file was touched. Separately, `Images.tsx` now sits at exactly 250 — no headroom left.
- **6 — Two mechanisms reset the visited-tab cache, and one of them is unreachable** (Low, new). `CuneiformFragment.tsx:104` adds `key={fragment.number}` to `<Images>`, which remounts the component whenever the fragment changes. `Images.tsx:166-179` *also* resets `visitedTabState` when `visited.fragmentNumber !== fragment.number` (line 168). With the `key` in place that branch cannot run in the application; the only thing that reaches it is `Images.tabs.test.tsx:122` ('resets visited media tabs when the fragment changes'), which renders `Images` directly and so bypasses the `key`. Keep one or the other.
- **7 — `withData`'s retry handler closes over a stale count** (Low, new). `onRetry={fullConfig.retry === true ? () => setRetryCount(retryCount + 1) : undefined}` (`withData.tsx:87-94`, the increment is on line 91). Two clicks dispatched against the same render commit both compute the same next value, so the second is a no-op and no refetch is triggered. `setRetryCount((count) => count + 1)` is the safe form.
- **8 — The page size has two sources of truth, and the fallback is wrong by construction** (Low, new). `SearchResult` receives `resultPageSize` as a prop from `FragmentariumSearch.tsx:114`, while `PaginationItems.tsx:88-90` independently re-derives it from the URL via `getValidatedPageSize(new URLSearchParams(location.search).get('limit'))`. Worse, `SearchResult`'s own fallback `resultPageSize ?? getValidatedPageSize(fragmentQuery.limit)` (`FragmentariumSearchResult.tsx:92-93`) can never be right for a line query, because `createPagedFragmentQuery` deliberately sends `limit + 1` (26 / 51 / 101) and none of those is in `RESULT_PAGE_SIZES`, so `getValidatedPageSize` silently returns 50. It only works today because `FragmentariumSearch` always passes the prop — and finding 1 is precisely what happens when a call site doesn't.
- **9 — `unknown` parameter types paper over a lying type** (Low). `getValidatedPageSize(value: unknown)` and `getPageIndexForOffset(offset: unknown, limit: unknown)` (`pagination.ts:73`, `:82`) breach the project rule against `unknown` unless very necessary. The underlying cause is that `FragmentQuery.limit` and `offset` are typed `number`, but `parseFragmentSearchQuery` (`fragmentariumRoutes.tsx:39-46`) produces them from `query-string.parse` without `parseNumbers`, so at runtime they are strings. Either widen the declared type to `number | string` or parse numbers at the boundary; then these helpers can take concrete types.
- **10 — `toWellFormed` is a hand-rolled polyfill of a standard method** (Low). `FragmentLink.tsx:6-25` reimplements `String.prototype.toWellFormed` (ES2024, supported by every browser this project targets). `value.toWellFormed?.() ?? value` would remove 20 lines. The behaviour itself is correct — I checked the surrogate-pair advance — and it is fully covered by `createFragmentUrl.test.tsx:78` and `:82`.
- **11 — `AnnotationsView.integration.test.tsx` no longer exercises routing** (Info). The old `.ts` version drove the whole app through `AppDriver` at `/library/Test.Fragment/annotate`; the rewrite renders `TagSignsView` directly. Replacing the 761-line snapshot with explicit role assertions is a clear improvement, but the route wiring for that path is no longer covered by this test. Worth a one-line route-level assertion somewhere if it isn't covered elsewhere.
- **12 — `head.test.tsx` mutates a global and never restores it** (Info). `HelmetProvider.canUseDOM = false` at `head.test.tsx:6` is module-scope. This is the round-1 finding 15 pattern relocated out of `FragmentView.test.tsx`; it is safe under Jest's per-file module registry, so no action is strictly required, but it is still an unrestored global. Also, two tests in that file share the name `'should render all metadata'` (lines 8 and 22) — pre-existing.
- **13 — Out-of-range `paginationIndex` is still not clamped** (Info). `?paginationIndex=9999` on `/library/search` produces `offset=499950`, an empty page, and a **Previous** button that steps to 9998. This is the deliberate consequence of removing the redirect that round-2 finding 9 objected to, and it is better than the silent bounce to page 1, but a user who lands there still has no practical way back. Clamping to the last known page when a page past the first comes back empty would close it.

khoidt (approved):
This one is basically there. Both blockers from last round got real answers. Nice work!

Two things left, and neither is really about the code you wrote.

The first is mechanical but hard: the branch **no longer merges**. `src/router/sitemap.test.tsx` conflicts, and GitHub reports `mergeable_state: dirty`. Master has moved a long way since your merge base — 201 files, +20,653/−11,855 — and some of it collides with this PR's assumptions: `CuneiformFragment.onSave` went back to Bluebird, and `SignImages.tsx` was gutted. This needs a merge and a full re-run of the gates afterwards, not just a conflict resolution.

The second is the carried-over one, unchanged: `GET /fragments/latest` on production still returns bare items — `museumNumber`, `matchingLines`, `matchCount` and nothing else. I checked again today. So the front page still fires 5 `find` calls and `/library` still fires one per item. The client half is correct and well tested; it's waiting on a backend change. I'd still like that confirmed before merge, since it's the PR's headline claim.

Approving, but please make sure that the merge is clean and the `/fragments/latest` question is answered. 🙂

Everything else is Low or Info — including one I verified by execution: the new repeated-retry test passes against the *old* buggy `withData`, so the fix is right but not pinned. Details below.

### Details

Numbering matches the `Findings` section further down, where each item repeats the evidence in full.

**Blockers**

- **1 — The branch does not merge into `master` any more** (Medium/High, new). The GitHub API reports `mergeable: false`, `mergeable_state: "dirty"`, and GraphQL `mergeable: CONFLICTING`. `git merge-tree --write-tree HEAD origin/master` produces one content conflict, in `src/router/sitemap.test.tsx` — master added `generateSitemapXml()` / `mockRealiaEntries()` helpers and three Realia sitemap tests in the same region this PR rewrote. Two more files auto-merge but are worth eyeballing: `src/fragmentarium/ui/fragment/CuneiformFragment.tsx` and `src/router/sitemap.tsx`. The bigger point is the divergence behind that one conflict — `git diff 4ac457d8..origin/master` is **201 files, +20,653 / −11,855**, and at least two of those changes interact with this PR: master changed `CuneiformFragment`'s `onSave` signature from `(updatedFragment: Promise<Fragment>) => void` back to `(updatedFragment: Bluebird<Fragment>) => Bluebird<Fragment>`, and master deleted 252 of the 262 changed lines in `src/signs/ui/display/SignImages.tsx` (the file carrying the pre-existing `tsc` errors in finding 3). This is not a "resolve the conflict and push" job: after the merge, `yarn lint`, `yarn tsc`, the full suite and the console-clean check all need re-running, because none of the green results below were measured against merged code.
- **2 — The Latest Additions fan-out removal is still inert against the production API** (High, carried from round 2 finding 2 / round 3 finding 2, unchanged). Re-probed today: `GET https://www.ebl.lmu.de/api/fragments/latest` returns top-level keys `["matchCountTotal","items"]` and 50 items whose only keys are `matchingLines`, `museumNumber`, `matchCount` — for example `{"matchingLines":[1,3,4],"museumNumber":{"prefix":"IM","number":"77167","suffix":""},"matchCount":0}`. No `fragment`, no `thumbnailPath`, no top-level `fragments` array. `createLatestQueryResult` (`FragmentRepository.ts:295-320`) can only attach a fragment from (a) a summary item — `isQuerySummaryItemDto` requires `'description' in dto && 'script' in dto && 'hasPhoto' in dto` (`FragmentRepository.ts:239`), none of which are present — or (b) a top-level `fragments` prefetch array the API does not send. So `queryItem.fragment` stays `undefined` for all 50 items, `hasRenderReadyFragment` is false, and both surfaces hydrate per card: the front page issues 5 `fragmentService.find` calls (`Introduction.tsx` → `preview={true}` → `LatestTransliterations.tsx` → `CompactFragmentCard` → `HydratedCompactFragmentCard`) and `/library` issues one per item via `includeLatestRecord={true}`. Even a *summary*-shaped response would not short-circuit `/library`, because `includeLatestRecord: true` additionally requires a non-historical `Transliteration` record (`queryItemRenderReady.ts:5-9`) and `createQuerySummaryFragment` hardcodes `record: []` (`FragmentRepository.ts:263`) — only the `fragments` array can satisfy that. The client code is correct and the mapper-driven tests are honest about it (`LatestTransliterations.mapper.test.tsx:147` asserts that a summary item *does* still hydrate). Per the project's own rule that the backend schema is the source of truth, shipping a client that reads a `fragments` field the API does not emit needs backend confirmation. Note the search-side equivalent **does** work: `GET /fragments/query?...` returns full summary items with `thumbnailPath`, so `/library/search` and the project pages genuinely avoid per-card hydration.

**Hard gates**

- **3 — `yarn tsc` is red** (Medium, **pre-existing — not introduced by this PR**). Two errors at head: `src/signs/ui/display/SignImages.tsx(297,41)` and `(306,42)`, both `TS18046: 'result.value' is of type 'unknown'`. `git diff --name-only 4ac457d8..HEAD | grep -i signimages` returns nothing — the file is not in this PR's diff — and the same two errors reproduce at the merge base. I also ran `tsc` against the **current** `origin/master` tip (`1f5b31c8`) in a throwaway worktree: master is red too, with a different single error, `src/fragmentarium/ui/text-annotation/realiaOptionLoader.ts(30,9): error TS18046: 'entries' is of type 'unknown'`. So merging master (finding 1) will very likely make the `SignImages` pair disappear — master deleted that code — and inherit the `realiaOptionLoader` one instead. The gate stays red either way and needs its own change. Root cause on `SignImages.tsx`: `runWithConcurrencyLimit<T, R>` declares `task: (item: T) => Promise<R>`, but the call site passes `signService.getClusterVariants(...)`, which returns a Bluebird promise; inference against the native `Promise<R>` shape fails and `R` collapses to `unknown`. Not cosmetic — `craco start` reports "Compiled successfully!" and then surfaces both as full-screen error-overlay entries.

**Everything else**

- **4 — `PdfExport.tsx` is still 776 lines** (Low, pre-existing). It was 775 on `master`; this PR nudged it to 776 while changing `getHyperLink`. It remains the *only* file this PR touches that exceeds the 250-line ceiling — I line-counted every changed `.ts`/`.tsx` against the merge base. Worth noting how close a few others now sit: `Images.tsx` 239, `researchProjectRoutes.pagination.test.tsx` 233, `chapter.ts` 232, `FragmentLemmaLines.tsx` 231. Flagging only because the ceiling is a hard gate and the file was touched.
- **5 — The new repeated-retry test passes against the un-fixed `withData`** (Low, new). The fix itself is right: `withData.tsx:88-92` now reads `onRetry={fullConfig.retry === true ? () => setRetryCount((currentRetryCount) => currentRetryCount + 1) : undefined}`. But `withData.retry.test.tsx:68-84` ('keeps re-querying the API on every repeated retry') clicks **Retry**, awaits the error, then clicks again — two clicks in two separate render commits, which the stale-closure form handles correctly too. I verified this by execution: I temporarily restored `() => setRetryCount(retryCount + 1)`, ran `yarn test --watchAll=false src/http/withData.retry.test.tsx`, and got **3 passed** — the test does not fail on the old code. (The file was restored immediately; `git diff` is empty.) To actually pin the fix the two clicks have to land in the same commit, e.g. `await act(async () => { retryButton.click(); retryButton.click() })`, asserting the getter is called three times.
- **6 — `toWellFormed` is a hand-rolled polyfill of a standard method** (Low, carried, unchanged). `FragmentLink.tsx:8-30` reimplements `String.prototype.toWellFormed` (ES2024, supported by every browser this project targets). `value.toWellFormed?.() ?? value` would remove ~20 lines. The behaviour is correct — I checked the surrogate-pair advance — and it is fully covered by `createFragmentUrl.test.tsx:78` and `:82`.
- **7 — The overfetch and the server's `hasNextPage` are two answers to the same question** (Info, new, latent). `FragmentariumSearchResult.tsx:94-95` reads `const effectiveHasNextPage = data.hasNextPage ?? data.items.length > pageSize`, i.e. the server's answer wins when present. For line queries `createPagedFragmentQuery` deliberately requests `limit: pageSize + 1` with `count: 'exact'`, and the live API returns `hasNextPage: null` for `count=exact` (confirmed: `lemmas=šarru I&limit=10&count=exact` → `matchCountTotal: 9020`, `hasNextPage: null`), so today the fallback always runs and is correct. If the backend ever starts populating `hasNextPage` for `count=exact`, it will be answering "are there more than `pageSize + 1` results?" while the client paginates in `pageSize`-wide pages — with exactly 51 results the server would say `false` and page 2, which holds one real item, would become unreachable. Cheap to make explicit: prefer the derived value whenever the request overfetched.
- **8 — `FragmentLemmaLines` now asks for an exact count, which makes its own fallback branch unreachable** (Info, new). `FragmentLemmaLines.tsx:229` changed `count: 'none'` → `count: 'exact'`. This is the better resolution of the Codex P2 thread — the page now reports a real total ("9,020 matches", "Show all 9,020 matches in Library search") instead of describing a capped 10-item sample. I checked the cost against production: `?lemmas=šarru%20I&limit=10&count=exact` took 5.27 s versus 5.72 s for `count=none`, so the exact count is lost in the noise of the query itself. The side effect is that `hasMatchCount` (`FragmentLemmaLines.tsx:176`) is now always true against the live API, leaving the `Showing N Library document examples` branch (`:193-205`) unreachable outside tests — the same dead-branch shape that round-3 finding 4 objected to, just harmless this time. Either drop it or leave a note that it exists for a backend that ignores `count`.
- **9 — `/library/:id/annotate` route wiring is still uncovered** (Info, carried from round 3 finding 11). The rewritten `AnnotationsView.integration.test.tsx` renders `TagSignsView` directly instead of driving the route through `AppDriver`, and nothing replaced the route-level coverage. This now shows up in the coverage report: `src/router/fragmentariumRoutes.tsx` line **129** — the `render` callback of the `TagSignsView` route — is listed as uncovered. The only other test that mentions the path is `notFoundRoutes.test.tsx:90`, which asserts that `/library/Fragment.12345/annotate/non-existent` renders NotFound, i.e. it exercises the *absence* of a route rather than this one. Replacing a 761-line snapshot with explicit role assertions was still the right call; it just left this gap.
- **10 — `head.test.tsx` mutates a global and never restores it** (Info, carried, unchanged). `HelmetProvider.canUseDOM = false` at `head.test.tsx:6` is module-scope. Safe under Jest's per-file module registry, so no action is strictly required, but it is still an unrestored global. Two tests in that file also share the name `'should render all metadata'` (lines 8 and 22) — pre-existing.
- **11 — Out-of-range `paginationIndex` is still not clamped** (Info, carried, unchanged, now with wider reach). `?paginationIndex=9999` on `/library/search` produces `offset=499950`, an empty page, and a **Previous** button that steps to 9998. Since finding 1 of round 3 was fixed by giving the research-project pages the same pagination, `/projects/CAIC/search?paginationIndex=9999` now behaves the same way. This is the deliberate consequence of removing the redirect that round-2 finding 9 objected to and is still better than the silent bounce to page 1, but clamping to the last known page when a page past the first comes back empty would close it.
- **12 — Dead defensive `omit`s left behind by the type split** (Info, new, nit). `hasNonDefaultValues` (`FragmentariumSearch.tsx:37-44`) still does `.omit('limit').omit('offset').omit('count')`, but the value it now receives is a `FragmentSearchCriteria`, which is defined as `Omit<FragmentQuery, 'limit' | 'offset' | 'count'>` (`FragmentQuery.ts:31-34`) and is built by `parseSearchCriteria`, which strips exactly those keys plus `paginationIndex`. The omits can only ever be no-ops for the fragment side; they still matter for the `CorpusQuery` branch of the union, so if you keep them a one-word note would help.
- **13 — `SummaryThumbnail` assumes the thumbnail path is always API-relative** (Info, new, nit). `SummaryThumbnail.tsx:23` now reads `src={apiUrl(thumbnailPath)}`, which is the right fix — the raw value resolved against the *frontend* origin before, so summary thumbnails were broken. Confirmed against production: `thumbnailPath` is `/fragments/IM.2899/thumbnail/small`, and `GET https://www.ebl.lmu.de/api/fragments/BM.108866/thumbnail/small` returns `200 image/jpeg` with **no** `Authorization` header, so a plain `` is correct and no blob fetch is needed. Two small things: `apiUrl` would corrupt an absolute URL if the backend ever emits one (`apiUrl` is a bare template concat, `ApiClient.ts:13-15`), and the API emits a `thumbnailPath` even for items with `hasPhoto: false` (all 50 CAIC items on page 1 are such) — harmless today only because `FragmentariumSearchResultComponents.tsx:174` gates the render on `fragment.hasPhoto && isNearViewport`.

Fabdulla1:
> This one is basically there. Both blockers from last round got real answers. Nice work!
> 
> Two things left, and neither is really about the code you wrote.
> 
> The first is mechanical but hard: the branch **no longer merges**. `src/router/sitemap.test.tsx` conflicts, and GitHub reports `mergeable_state: dirty`. Master has moved a long way since your merge base — 201 files, +20,653/−11,855 — and some of it collides with this PR's assumptions: `CuneiformFragment.onSave` went back to Bluebird, and `SignImages.tsx` was gutted. This needs a merge and a full re-run of the gates afterwards, not just a conflict resolution.
> 
> The second is the carried-over one, unchanged: `GET /fragments/latest` on production still returns bare items — `museumNumber`, `matchingLines`, `matchCount` and nothing else. I checked again today. So the front page still fires 5 `find` calls and `/library` still fires one per item. The client half is correct and well tested; it's waiting on a backend change. I'd still like that confirmed before merge, since it's the PR's headline claim.
> 
> Approving, but please make sure that the merge is clean and the `/fragments/latest` question is answered. 🙂
> 
> Everything else is Low or Info — including one I verified by execution: the new repeated-retry test passes against the _old_ buggy `withData`, so the fix is right but not pinned. Details below.
> 
> ### Details
> Numbering matches the `Findings` section further down, where each item repeats the evidence in full.
> 
> **Blockers**
> 
> * **1 — The branch does not merge into `master` any more** (Medium/High, new). The GitHub API reports `mergeable: false`, `mergeable_state: "dirty"`, and GraphQL `mergeable: CONFLICTING`. `git merge-tree --write-tree HEAD origin/master` produces one content conflict, in `src/router/sitemap.test.tsx` — master added `generateSitemapXml()` / `mockRealiaEntries()` helpers and three Realia sitemap tests in the same region this PR rewrote. Two more files auto-merge but are worth eyeballing: `src/fragmentarium/ui/fragment/CuneiformFragment.tsx` and `src/router/sitemap.tsx`. The bigger point is the divergence behind that one conflict — `git diff 4ac457d8..origin/master` is **201 files, +20,653 / −11,855**, and at least two of those changes interact with this PR: master changed `CuneiformFragment`'s `onSave` signature from `(updatedFragment: Promise<Fragment>) => void` back to `(updatedFragment: Bluebird<Fragment>) => Bluebird<Fragment>`, and master deleted 252 of the 262 changed lines in `src/signs/ui/display/SignImages.tsx` (the file carrying the pre-existing `tsc` errors in finding 3). This is not a "resolve the conflict and push" job: after the merge, `yarn lint`, `yarn tsc`, the full suite and the console-clean check all need re-running, because none of the green results below were measured against merged code.
> * **2 — The Latest Additions fan-out removal is still inert against the production API** (High, carried from round 2 finding 2 / round 3 finding 2, unchanged). Re-probed today: `GET https://www.ebl.lmu.de/api/fragments/latest` returns top-level keys `["matchCountTotal","items"]` and 50 items whose only keys are `matchingLines`, `museumNumber`, `matchCount` — for example `{"matchingLines":[1,3,4],"museumNumber":{"prefix":"IM","number":"77167","suffix":""},"matchCount":0}`. No `fragment`, no `thumbnailPath`, no top-level `fragments` array. `createLatestQueryResult` (`FragmentRepository.ts:295-320`) can only attach a fragment from (a) a summary item — `isQuerySummaryItemDto` requires `'description' in dto && 'script' in dto && 'hasPhoto' in dto` (`FragmentRepository.ts:239`), none of which are present — or (b) a top-level `fragments` prefetch array the API does not send. So `queryItem.fragment` stays `undefined` for all 50 items, `hasRenderReadyFragment` is false, and both surfaces hydrate per card: the front page issues 5 `fragmentService.find` calls (`Introduction.tsx` → `preview={true}` → `LatestTransliterations.tsx` → `CompactFragmentCard` → `HydratedCompactFragmentCard`) and `/library` issues one per item via `includeLatestRecord={true}`. Even a _summary_-shaped response would not short-circuit `/library`, because `includeLatestRecord: true` additionally requires a non-historical `Transliteration` record (`queryItemRenderReady.ts:5-9`) and `createQuerySummaryFragment` hardcodes `record: []` (`FragmentRepository.ts:263`) — only the `fragments` array can satisfy that. The client code is correct and the mapper-driven tests are honest about it (`LatestTransliterations.mapper.test.tsx:147` asserts that a summary item _does_ still hydrate). Per the project's own rule that the backend schema is the source of truth, shipping a client that reads a `fragments` field the API does not emit needs backend confirmation. Note the search-side equivalent **does** work: `GET /fragments/query?...` returns full summary items with `thumbnailPath`, so `/library/search` and the project pages genuinely avoid per-card hydration.
> 
> **Hard gates**
> 
> * **3 — `yarn tsc` is red** (Medium, **pre-existing — not introduced by this PR**). Two errors at head: `src/signs/ui/display/SignImages.tsx(297,41)` and `(306,42)`, both `TS18046: 'result.value' is of type 'unknown'`. `git diff --name-only 4ac457d8..HEAD | grep -i signimages` returns nothing — the file is not in this PR's diff — and the same two errors reproduce at the merge base. I also ran `tsc` against the **current** `origin/master` tip (`1f5b31c8`) in a throwaway worktree: master is red too, with a different single error, `src/fragmentarium/ui/text-annotation/realiaOptionLoader.ts(30,9): error TS18046: 'entries' is of type 'unknown'`. So merging master (finding 1) will very likely make the `SignImages` pair disappear — master deleted that code — and inherit the `realiaOptionLoader` one instead. The gate stays red either way and needs its own change. Root cause on `SignImages.tsx`: `runWithConcurrencyLimit<T, R>` declares `task: (item: T) => Promise<R>`, but the call site passes `signService.getClusterVariants(...)`, which returns a Bluebird promise; inference against the native `Promise<R>` shape fails and `R` collapses to `unknown`. Not cosmetic — `craco start` reports "Compiled successfully!" and then surfaces both as full-screen error-overlay entries.
> 
> **Everything else**
> 
> * **4 — `PdfExport.tsx` is still 776 lines** (Low, pre-existing). It was 775 on `master`; this PR nudged it to 776 while changing `getHyperLink`. It remains the _only_ file this PR touches that exceeds the 250-line ceiling — I line-counted every changed `.ts`/`.tsx` against the merge base. Worth noting how close a few others now sit: `Images.tsx` 239, `researchProjectRoutes.pagination.test.tsx` 233, `chapter.ts` 232, `FragmentLemmaLines.tsx` 231. Flagging only because the ceiling is a hard gate and the file was touched.
> * **5 — The new repeated-retry test passes against the un-fixed `withData`** (Low, new). The fix itself is right: `withData.tsx:88-92` now reads `onRetry={fullConfig.retry === true ? () => setRetryCount((currentRetryCount) => currentRetryCount + 1) : undefined}`. But `withData.retry.test.tsx:68-84` ('keeps re-querying the API on every repeated retry') clicks **Retry**, awaits the error, then clicks again — two clicks in two separate render commits, which the stale-closure form handles correctly too. I verified this by execution: I temporarily restored `() => setRetryCount(retryCount + 1)`, ran `yarn test --watchAll=false src/http/withData.retry.test.tsx`, and got **3 passed** — the test does not fail on the old code. (The file was restored immediately; `git diff` is empty.) To actually pin the fix the two clicks have to land in the same commit, e.g. `await act(async () => { retryButton.click(); retryButton.click() })`, asserting the getter is called three times.
> * **6 — `toWellFormed` is a hand-rolled polyfill of a standard method** (Low, carried, unchanged). `FragmentLink.tsx:8-30` reimplements `String.prototype.toWellFormed` (ES2024, supported by every browser this project targets). `value.toWellFormed?.() ?? value` would remove ~20 lines. The behaviour is correct — I checked the surrogate-pair advance — and it is fully covered by `createFragmentUrl.test.tsx:78` and `:82`.
> * **7 — The overfetch and the server's `hasNextPage` are two answers to the same question** (Info, new, latent). `FragmentariumSearchResult.tsx:94-95` reads `const effectiveHasNextPage = data.hasNextPage ?? data.items.length > pageSize`, i.e. the server's answer wins when present. For line queries `createPagedFragmentQuery` deliberately requests `limit: pageSize + 1` with `count: 'exact'`, and the live API returns `hasNextPage: null` for `count=exact` (confirmed: `lemmas=šarru I&limit=10&count=exact` → `matchCountTotal: 9020`, `hasNextPage: null`), so today the fallback always runs and is correct. If the backend ever starts populating `hasNextPage` for `count=exact`, it will be answering "are there more than `pageSize + 1` results?" while the client paginates in `pageSize`-wide pages — with exactly 51 results the server would say `false` and page 2, which holds one real item, would become unreachable. Cheap to make explicit: prefer the derived value whenever the request overfetched.
> * **8 — `FragmentLemmaLines` now asks for an exact count, which makes its own fallback branch unreachable** (Info, new). `FragmentLemmaLines.tsx:229` changed `count: 'none'` → `count: 'exact'`. This is the better resolution of the Codex P2 thread — the page now reports a real total ("9,020 matches", "Show all 9,020 matches in Library search") instead of describing a capped 10-item sample. I checked the cost against production: `?lemmas=šarru%20I&limit=10&count=exact` took 5.27 s versus 5.72 s for `count=none`, so the exact count is lost in the noise of the query itself. The side effect is that `hasMatchCount` (`FragmentLemmaLines.tsx:176`) is now always true against the live API, leaving the `Showing N Library document examples` branch (`:193-205`) unreachable outside tests — the same dead-branch shape that round-3 finding 4 objected to, just harmless this time. Either drop it or leave a note that it exists for a backend that ignores `count`.
> * **9 — `/library/:id/annotate` route wiring is still uncovered** (Info, carried from round 3 finding 11). The rewritten `AnnotationsView.integration.test.tsx` renders `TagSignsView` directly instead of driving the route through `AppDriver`, and nothing replaced the route-level coverage. This now shows up in the coverage report: `src/router/fragmentariumRoutes.tsx` line **129** — the `render` callback of the `TagSignsView` route — is listed as uncovered. The only other test that mentions the path is `notFoundRoutes.test.tsx:90`, which asserts that `/library/Fragment.12345/annotate/non-existent` renders NotFound, i.e. it exercises the _absence_ of a route rather than this one. Replacing a 761-line snapshot with explicit role assertions was still the right call; it just left this gap.
> * **10 — `head.test.tsx` mutates a global and never restores it** (Info, carried, unchanged). `HelmetProvider.canUseDOM = false` at `head.test.tsx:6` is module-scope. Safe under Jest's per-file module registry, so no action is strictly required, but it is still an unrestored global. Two tests in that file also share the name `'should render all metadata'` (lines 8 and 22) — pre-existing.
> * **11 — Out-of-range `paginationIndex` is still not clamped** (Info, carried, unchanged, now with wider reach). `?paginationIndex=9999` on `/library/search` produces `offset=499950`, an empty page, and a **Previous** button that steps to 9998. Since finding 1 of round 3 was fixed by giving the research-project pages the same pagination, `/projects/CAIC/search?paginationIndex=9999` now behaves the same way. This is the deliberate consequence of removing the redirect that round-2 finding 9 objected to and is still better than the silent bounce to page 1, but clamping to the last known page when a page past the first comes back empty would close it.
> * **12 — Dead defensive `omit`s left behind by the type split** (Info, new, nit). `hasNonDefaultValues` (`FragmentariumSearch.tsx:37-44`) still does `.omit('limit').omit('offset').omit('count')`, but the value it now receives is a `FragmentSearchCriteria`, which is defined as `Omit<FragmentQuery, 'limit' | 'offset' | 'count'>` (`FragmentQuery.ts:31-34`) and is built by `parseSearchCriteria`, which strips exactly those keys plus `paginationIndex`. The omits can only ever be no-ops for the fragment side; they still matter for the `CorpusQuery` branch of the union, so if you keep them a one-word note would help.
> * **13 — `SummaryThumbnail` assumes the thumbnail path is always API-relative** (Info, new, nit). `SummaryThumbnail.tsx:23` now reads `src={apiUrl(thumbnailPath)}`, which is the right fix — the raw value resolved against the _frontend_ origin before, so summary thumbnails were broken. Confirmed against production: `thumbnailPath` is `/fragments/IM.2899/thumbnail/small`, and `GET https://www.ebl.lmu.de/api/fragments/BM.108866/thumbnail/small` returns `200 image/jpeg` with **no** `Authorization` header, so a plain `` is correct and no blob fetch is needed. Two small things: `apiUrl` would corrupt an absolute URL if the backend ever emits one (`apiUrl` is a bare template concat, `ApiClient.ts:13-15`), and the API emits a `thumbnailPath` even for items with `hasPhoto: false` (all 50 CAIC items on page 1 are such) — harmless today only because `FragmentariumSearchResultComponents.tsx:174` gates the render on `fragment.hasPhoto && isNearViewport`.

So production `/fragments/latest` still returns bare items, which means Latest Additions continues making per-item find requests. This PR only prepares the client to reuse prefetched fragments once the backend provides them. I updated the PR description to make that dependency explicit for that. Thanks for the review!

PR #771, opened 2026-07-14
fix(test): make fragmentInfoFactory numbers unique (flaky FragmentList duplicate key)

khoidt:
## Problem

`src/fragmentarium/ui/FragmentList.test.tsx` fails **intermittently** on `master`:

```text
FAIL src/fragmentarium/ui/FragmentList.test.tsx
  ● Console
    console.error
      Warning: Encountered two children with the same key, `ma`. Keys should be unique so that
      components maintain their identity across updates.

  ● No config › Fragment 1 › Links to the fragment
    TestingLibraryElementError: Found multiple elements with the text: ma
```

It passed on one run of a given commit and failed on the next, with no code change in between.

## Root cause

The component is not at fault — the **test fixture** is.

1. `FragmentList.tsx` keys its rows by the fragment number (correct — that is the fragment's identity):
   `<tr key={fragment.number}>`
2. `fragmentInfoFactory` generated `number: defaultChance.word()`.
3. `defaultChance = new Chance()` is **unseeded**, and `chance.word()` draws from a small space, so
   short words such as `ma` collide.

When two fragments in one list draw the same word, React sees duplicate keys **and**
`screen.getByText(fragment.number)` matches two `<a>` elements. The factory was silently violating an
invariant the component legitimately depends on: fragment numbers are unique.

## Fix

Make the number unique by construction, using the factory's monotonically increasing `sequence`:

```diff
 export const fragmentInfoFactory = Factory.define<FragmentInfo>(
-  ({ associations }) => ({
-    number: defaultChance.word(),
+  ({ associations, sequence }) => ({
+    number: `${defaultChance.word()}.${sequence}`,
```

The value keeps its random, museum-number-like shape (`ma.17`) while guaranteeing distinctness for
any seed.

Notably **not** done: silencing `console.error`, or loosening the assertion to `getAllByText(...)[0]`.
Both would hide the warning while leaving the invalid render in place.

## Evidence

Repeatedly running the flaky test and seeing it pass proves little for a ~1-in-a-few-hundred failure,
so the invariant is measured directly — build 300 fragments, count distinct numbers:

| Factory | Fragments built | Distinct numbers |
| --- | --- | --- |
| before (`chance.word()`) | 300 | **291** — 9 collisions |
| after (word + sequence) | 300 | **300** |

## Verification

| Check | Result |
| --- | --- |
| `FragmentList.test.tsx` ×10 | 15/15 tests pass every run, zero console output |
| `yarn lint` | clean |
| `yarn tsc` | clean |
| `yarn test --watchAll=false` | 339/339 suites, 3460 passed, 2 skipped, 0 failed, zero console output |

## Documentation

Adds `docs/flaky-test-fragment-list-duplicate-key.md` describing the symptom, root cause, evidence,
the fix, the alternatives rejected and why, and how to recognise the same class of bug in other
factories (any generated field used as a React key / lookup id / URL segment must be unique per build).

---

Found while reviewing #765 — this failure is **pre-existing and unrelated** to that PR, but it blocks
a clean, console-silent suite run for every PR until fixed.

### DIFF
--- a/docs/flaky-test-fragment-list-duplicate-key.md
+++ b/docs/flaky-test-fragment-list-duplicate-key.md
@@ -1,0 +1,129 @@
+# Flaky test: `FragmentList` duplicate React key
+
+## Summary
+
+`src/fragmentarium/ui/FragmentList.test.tsx` failed intermittently on `master` with a React
+duplicate-key `console.error` and an ambiguous element query. The cause was test **fixture data**,
+not the component: `fragmentInfoFactory` generated fragment numbers from an unseeded random word
+list, so two fragments in the same list occasionally received the same `number` — which
+`FragmentList` uses as its React key.
+
+This document records the issue, the evidence, and the fix, so the same class of bug is recognisable
+if it appears in another factory.
+
+## Symptom
+
+The failure is intermittent: the suite passed on one run of a given commit and failed on the next,
+with no code change in between.
+
+```text
+FAIL src/fragmentarium/ui/FragmentList.test.tsx
+  ● Console
+    console.error
+      Warning: Encountered two children with the same key, `ma`. Keys should be unique so that
+      components maintain their identity across updates. Non-unique keys may cause children to be
+      duplicated and/or omitted — the behavior is unsupported and could change in a future version.
+          at tbody
+          at table
+          at FragmentList (src/fragmentarium/ui/FragmentList.tsx:14:3)
+
+  ● No config › Fragment 1 › Links to the fragment
+    TestingLibraryElementError: Found multiple elements with the text: ma
+```
+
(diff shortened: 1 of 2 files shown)


Fabdulla1 (approved):
Everything looks good except 2 non-blocking issues. One being, I think it would be great to add a regression test. The sequence suffix makes the default-generated fragment numbers unique by construction. I think a regression test here would directly protect the invariant and prevent a future refactor from accidentally reintroducing the flaky duplicate-key behavior. The second issue is that the documentation reports 291 distinct values and nine collisions from 300 generated fragments. Since `defaultChance` is unseeded, the exact result will vary between runs. It might be beneficial to describe this as the outcome of one observed run rather than a fixed expected count?

PR #773, opened 2026-07-21
chore: migrate TypeScript config off TS7-deprecated compiler options

khoidt:
## Summary

Migrates the TypeScript compiler config off the options TypeScript 7 will remove, replacing each with a non-deprecated equivalent — a **real** migration, no `ignoreDeprecations` suppression left behind. Behaviour is preserved throughout.

## Before / after `tsconfig.json`

| Option | Before | After |
| --- | --- | --- |
| `target` | `es5` | `es2020` |
| `moduleResolution` | `node` (node10) | `bundler` |
| `baseUrl` | `src` | removed |
| `paths` | — | `{ "*": ["./src/*"] }` |
| `downlevelIteration` | `true` | removed |
| `ignoreDeprecations` | `"5.0"` | removed |

`paths` targets must be relative when `baseUrl` is unset (`TS5090`), hence `./src/*`.

## CRACO compensation

CRA (`react-scripts` 5.0.1) derives absolute-import resolution from tsconfig `baseUrl`, **not** `paths` (`react-scripts/config/modules.js`). Removing `baseUrl` therefore requires restoring `src` to the webpack and jest resolvers in `craco.config.js`:
- webpack `resolve.modules` now includes `src`;
- jest `modulePaths` now includes `src` (new `jest.configure` block).

(eslint-plugin-import resolves absolute imports independently via `.eslintrc` `moduleDirectory`, already `src/` — lint is unaffected.)

## Migration breakages fixed (root cause, behaviour-preserving)

- **dotenv typing** (`setupTests.ts`): under `bundler`, `dotenv.parse()` values type as `unknown` → `String(value)`.
- **async return types (`TS1064`, 51 errors / 23 files):** raising `target` off `es5` makes TS enforce that an `async` function's return annotation is the **global** `Promise`. These files annotated async functions with a shadowed `Promise` (`import Promise from 'bluebird'`) or `Bluebird<T>` — type-fictions, since an `async` function always returns a native Promise at runtime. Corrected to the global `Promise`; vestigial bluebird imports dropped. Test mocks feeding Bluebird-typed services keep explicit `Bluebird.resolve/reject` (restores exact prior runtime behaviour).
- **Genuine Bluebird preserved:** `withData` (`.cancel`/`.isCancelled`) and the `usePromiseEffect` date chain stay on Bluebird. `sitemap`'s `withData` getter bridges via `Bluebird.resolve(...)`; an `await`-free function was de-asynced rather than re-typed; `Content`/`DateSelection` prop types (only `.then`) were widened `Bluebird→Promise`.

## Risk notes

- `target es2020` is type-check-only (`noEmit`); the actual bundle transpile target is driven by `browserslist` via Babel and is unchanged. `browserslist` already excludes IE11 / Safari <12 / iOS ≤14.7 — well within es2020.
- `bundler` honours `package.json` `exports`; only `dotenv` surfaced a typing difference, fixed.
- Bluebird remains a dependency (160+ files use it for cancellation); this PR only corrected async **return-type** annotations — it does not remove Bluebird.

## Verification

- `yarn tsc` → clean (exit 0)
- `yarn lint` → clean (exit 0)
- `CI=true yarn test --watch=false` → **340 suites, 3469 passed / 2 skipped (pre-existing), 50 snapshots, zero console output**

## Note

This branch includes the working-tracking docs `TASK-ts7-migration-{research,todo,log}.md`. **Remove them before merge.**


Fabdulla1 (approved):
Excited for the update B-) 
Did a full review and everything looks great!

khoidt:
-> PR #774

PR #774, opened 2026-07-21
chore: remove bluebird, use AbortController for cancellation

khoidt:
## Summary

Removes the **`bluebird`** dependency, replacing its promise cancellation with the web-standard **`AbortController`/`AbortSignal`**. The `AbortController` lives in the React layer and a `signal` threads down through the fetch chain.

> Stacked on #773 (TS7 tsconfig migration) — this PR's base is `chore/ts7-tsconfig-migration`, so it shows only the bluebird diff. GitHub will retarget it to `master` when #773 merges.

## Why AbortSignal threading

Bluebird was used almost entirely for **cancellable promises** (`.cancel()`/`.isCancelled()`/`onCancel`). `cancellableFetch` already used an `AbortController` internally — bluebird just wrapped it to expose `.cancel()`. Threading a native `AbortSignal` removes the dependency without a custom promise abstraction.

## Core changes

| Area | Change |
| --- | --- |
| `cancellableFetch` | native `fetch(url, options)` (signal via options) |
| `ApiClient` | native promises; `signal?` threaded through every method into fetch options; `AbortError` excluded from error capture |
| `withData` | owns an `AbortController`, passes `signal` to the getter, `abort()`s on cleanup; cancellation via `AbortError`/`signal.aborted` |
| `usePromiseEffect` | new `[run, cancel]` `AbortController` API; 8 consumers migrated |
| `ConcurrencyLimiter` | native queue; optional `signal` aborts a *queued* op |
| `getOrFetchCachedValue` | in-flight requests reused; dropped `.isCancelled()` cache eviction |
| save-forms + `AfoRegisterSearchForm` | `AbortController` field/ref; gate `setState` on `signal.aborted` |

## Bluebird-specific helpers

- `mapSeries` → new `common/utils/mapSeries` (TextService, LemmatizationFactory)
- `Bluebird.map({ concurrency })` → `ConcurrencyLimiter` (CuneiformConverterForm)
- `Promise.config` removed (index.tsx, setupTests.ts)
- `.all(promise-of-array)` deep-resolve simplified (AfoRegisterRepository, GlossaryFactory, WordDisplayLogograms)

## Behaviour notes (deliberate)

- Data fetches abort the network on unmount **iff** the getter threads the signal; where it doesn't, `withData`'s existing `requestSequence` guard still prevents stale state — no correctness regression.
- Save/download flows gate `setState` on `signal.aborted` rather than aborting an in-flight write.
- A "cancelled in-flight request" is no longer a concept; the request cache reuses in-flight requests and evicts on settle.
- `bluebird` + `@types/bluebird` removed from `package.json` (a transitive copy remains under other packages; src has zero references).

## Cancellation test suites

Rewritten from `.cancel()`/`.isCancelled()` assertions to `AbortController`-observable behaviour (signal aborted, no state update after unmount, signal forwarded to fetch, aborted-queued-op removed, re-fetch after failed in-flight). One WordDisplay snapshot updated (it had captured an error-alert artifact from an unmocked service that bluebird's lenient `.all(undefined)` hid).

## Verification

- `yarn tsc` → clean
- `yarn lint` → clean
- `CI=true yarn test --watch=false` → **340 suites, 3470 passed / 2 skipped, 50 snapshots, zero console output**
- `yarn build` → (result in PR thread)

## Note

Includes the working-tracking docs `TASK-remove-bluebird-{todo,log}.md`. **Remove before merge.**


Fabdulla1 (changes requested):
Time to get bluebird out of here! It's great work but there are some concerns I have that I'll address below.

First, I noticed that `runWrite` can abort an already dispatched server write. `AbortableOperation.start()` in `src/common/utils/AbortableOperation.ts:4-8` always aborts the previous controller before creating a new one. `usePromiseEffect.runWrite()` uses that same supersession behavior in `src/common/hooks/usePromiseEffect.ts:10-23,39-43`. Because the signal is passed all the way into the write request, I can see that starting a second save can abort the first write's underlying `fetch`. One confirmed path is:

DateSelectionMethods.ts:43-59
→ FragmentService.ts:347-356
→ FragmentRepository.ts:522-531
→ ApiClient.postJson
→ fetch

The same pattern is reachable from:

ChapterEditView.tsx:59
CuneiformFragment.tsx:145-173
ScriptSelection.tsx:63

If the first request has already reached or been processed by the server, the client can no longer know its outcome after aborting the connection, while the abort is silently treated as cancellation. This conflicts with the stated goal of keeping writes running.

Could `runWrite` avoid network-aborting an already-dispatched write and instead only suppress stale UI updates? Alternatively, every consumer would need a reliable guard proving that overlapping writes are impossible.

Another concern I had was that `src/common/hooks/usePromiseEffect.test.tsx:52-58` runs the same “abort the previous operation when superseded” expectation for both run and runWrite.

That test confirms the current implementation, but it does not prove the desired write guarantee. Once the write-cancellation design is resolved, please add an integration-level test that reaches a mocked ApiClient or fetch and verifies that starting a second write does not network-abort a write that has already been dispatched.

The test should separately prove that stale results from the older write cannot update current UI state.

There are also some files that hit the 250-line ceiling: 
`src/fragmentarium/application/FragmentService.ts` — 888 lines
`src/fragmentarium/infrastructure/FragmentRepository.ts` — 787 lines
`src/corpus/application/TextService.ts` — 597 lines
`src/test-support/FakeApi.ts` — 516 lines
`src/signs/ui/display/SignImages.tsx` — 442 lines
`src/realia/ui/Realia.sass` — 453 lines
`src/http/withData.test.tsx` — 264 lines

Fabdulla1 (changes requested):
So great work on this and it's come a long way! There are some concerns I have listed below:
So it seems the PR still targets a closed and unmerged base. I noticed it targets `chore/ts7-tsconfig-migration` but that was closed without merging and its head is not an ancenstor of master. Merging #774 as currently configured would update that closed branch rather than land this work on master. 

Another thing that came up was that the bibliography cache can cross authentication scopes. `clear()` is called when the authentication cache scope changes, but an in-flight batch that settles afterward unconditionally repopulates `cachedEntries` here. Repro: user A starts `findMany`, the session changes to user B and clears the loader, then A's response resolves. B can subsequently receive A's cached entry without another request. The current batch test explicitly expects this unsafe behavior, while the single-entry path correctly prevents caching after a clear. Please guard batch and fallback cache writes with a generation/request identity and invert that regression test.

Another thing that I also noticed was that the fragment sidebar allows overlapping non-abortable writes. Serialize writes instead of only superseding their callbacks. `handleSave` dispatches every save, while `saving` disables only `EditorTabs`; `Info` and `GenreEditor` remain interactive. Delay an `updateGenres` add, then delete the genre before the first request settles and resolve the delete before the add: both POSTs reach the backend, the older add can be applied last, while `applyWhenCurrent` hides its response and leaves the UI showing the delete. Since writes are deliberately non-abortable, it might be good to just disable all fragment write controls while a save is pending or queue/serialize the writes, and cover reverse completion order.

Also, date Save/Delete remain active during pending writes. I think it would be good to prevent overlapping date writes while `isSaving`. The pending state is rendered, but Save ignores it and Delete is explicitly always enabled. A user can submit date A, change or delete it, and submit B before A settles; if B completes first and A last, the stale UI callback is hidden but the backend and fragment cache can finish on A. `DatesInTextSelection` also bypasses `runWrite`, and list-row editors do not share the parent's pending lock. Please disable the inputs and Save/Delete/Add controls or serialize these writes, and add reverse-resolution tests for both date paths.

Also, limiter can now lose cancellation during slot handoff. Would you be able to please re-check `signal.aborted` after `await acquireSlot(signal)` and before invoking `operation()`. Slot handoff seems to remove the abort listener and resolves the acquisition promise, but `run()` resumes in a later microtask. An abort in that gap is lost and the canceled queued operation executes. Maybe we keep the check inside `try/finally` so the granted slot is released, and strengthen the handoff-race test to assert that the queued operation was not called and rejected with the abort reason.

I also noticed that three `withData` callers still drop available signals. The round-8 forwarding repair is incomplete. Three `withData` getters still ignore the controller signal even though their service/repository paths already accept one:

- `Annotator.tsx`: pass the signal to `fragmentService.findPhoto(fragment, signal)`.
- `SearchFormPeriod.tsx`: pass it to `fragmentService.fetchPeriods(signal)`.
- `WordDisplayLogograms.tsx`: pass it to `signService.search(query, signal)`.

These reads were cancellable through the returned Bluebird chain before this migration; they now continue after navigation or unmount. Please forward the signal in all three getters and add caller-to-fetch assertions.

Converter cancellation never reaches HTTP either. We could thread this signal into the actual conversion fetch. `conversionOperation.start()` aborts the previous signal, but `query(line)` drops it. `SignService.getUnicodeFromAtf` and `SignRepository.getUnicodeFromAtf` likewise expose no signal and call `fetchJson` without one. With four slow lines, clicking Convert again or unmounting leaves all four old HTTP requests running. I noticed only queued work and stale rendering are canceled. Please pass `signal` through query → service → repository → `ApiClient.fetchJson`, and assert the real `RequestInit.signal`.

Dossier cancellation is converted into success as well. This catch also handles the `AbortError` produced by the newly threaded signal. Navigating away therefore logs an expected cancellation as “Failed to fetch dossiers” and converts it to a successful empty result, preventing `withData` from taking its cancellation path and allowing a post-unmount `setData` attempt. Please rethrow aborts with `isAbortError` before the graceful fallback, and add an abort-specific repository test.

Now, write callbacks remain current after unmount. Keeping an already-dispatched write alive is reasonable, but its UI continuation should not remain current after the hook unmounts. All current `runWrite` callers use `applyWhenCurrent` handlers that call React setters, while cleanup aborts only `readOperation`; a post-navigation completion therefore still invokes those handlers. Please call `writeOperation.current.supersede()` during cleanup without aborting the request, and update the unmount test to expect the write token to be stale.

PR #781, opened 2026-08-06
Add new manuscript types

ejimsan:


Fabdulla1 (changes requested):
So double checked both backend and frontend. Currently, the backend doesn't serialize manuscript type using PascalCase identifiers. `ManuscriptSchema.type` uses `ManuscriptType.long_name` for both serialization and deserialization. Basically, the backend wire values for these new types are:

- Multi-column tablet
- Collective tablet
- Student-teacher tablet
- School lentils
- Prisms
- Uncertain
But this PR uses these name values instead, which `toManuscriptDto()` sends directly:

- MultiColumnTablet
- CollectiveTablet
- StudentTeacherTablet
- SchoolLentil
- Prism
- Uncertain
It would be good to update the frontend API facing names to match the backend `long_name` strings exactly while keeping any internal keys/identifiers separate if needed.

Because of this, I also think it might be good to add a round-trip test for the new manuscript types as well. Just in case they ever get affected or changed. Adding a test that independently states the expected backend wire strings and verifies all six new types round-trip through the actual DTO mapper would be good.

Fabdulla1 (approved):

PR #787, opened 2026-08-12
Fix n calls

Fabdulla1:
This PR reduces frontend API fan-out by moving Fragmentarium and research-project searches to bounded server-side pagination and render-ready summary responses, while keeping dictionary fragment queries limited and avoiding per-card detail and thumbnail hydration. It also restores card parity for summary-backed results by rendering matching-line previews, correctly mapping script periods, preserving bibliography reference identities with honest fallbacks, and improving pagination visibility and responsive alignment without reintroducing N+1 requests. Latest Additions remains prepared to consume prefetched fragment data, but its remaining per-item requests still depend on a backend `/fragments/latest` enhancement.


khoidt (changes requested):
Nice piece of work — this is a big, genuinely useful change and it lands well. The N+1 fan-out is really gone, the summary-card path degrades gracefully instead of blowing up, the giant snapshot files are replaced with assertions that actually say something, and the whole suite runs green and completely silent (4160 tests, zero console noise). That last part is rarer than it should be, so thank you.

A handful of things I'd like sorted before this goes in. Two are quick: there are two explanatory comments in the diff that the project style rule doesn't allow, and the "latest transliteration record" predicate now exists in three places with two different definitions — one of them will show a historical record on a card that was admitted on the strength of a non-historical one. Two are worth a real look: a summary item with a malformed `museumNumber` still takes down the entire results page rather than falling back to the unavailable-card you built for exactly this, and for non-line queries there is no client-side way to detect a next page at all, so if the backend ever omits `hasNextPage` the Next button is dead forever. The thumbnail link also quietly changed from "new tab" to "full page reload in this tab".

Separately, and not your fault: `yarn tsc` is currently failing on `master` and therefore here too. I verified it on a clean `origin/master` worktree — same single error. Flagging it because the project treats the type gate as hard, and because CI doesn't run `tsc`, which is why it slipped through in the first place.

Also, as requested — good news on both counts: **no new `.md` files** and **no dev container / Docker / CI / build configuration changes** anywhere in this PR. Nothing to warn you about there.

### Details

#### B1 — `yarn tsc` fails (pre-existing on `master`, blocks the TypeScript hard gate)

**Severity:** Blocker · **Origin:** Pre-existing on `master`, not introduced by this PR

```
src/fragmentarium/ui/text-annotation/realiaOptionLoader.ts(30,9): error TS18046: 'entries' is of type 'unknown'.
```

The file is not in this PR's diff. It arrived on `master` via `8971a666` ("Add realia annotation layer to named entity annotation", #767) and is byte-identical between this branch and `origin/master`. I confirmed the failure is not an interaction with this PR by creating a detached worktree at `origin/master`, symlinking `node_modules`, and running `tsc` there — it produces the same single error.

Mechanical cause, in `loadRealiaOptions` (`src/fragmentarium/ui/text-annotation/realiaOptionLoader.ts:22-35`): the helper `toNativePromise<T>(promise: PromiseLike<T>): Promise<T>` fails to infer `T` from `RealiaService.search()`'s Bluebird `Promise<readonly RealiaEntry[]>` (`src/realia/application/RealiaService.ts:44`), so the `.then((entries) => …)` parameter widens to `unknown`.

Deeper cause worth fixing at the same time: **CI does not run `yarn tsc`.** The `test` check on this PR is green while the type gate is red. Until `tsc` is a CI check, type errors will keep reaching `master`.

#### F1 — "Latest transliteration record" logic duplicated three times, with two different predicates

**Severity:** High · **Files:** `src/query/queryItemRenderReady.ts:8-11`, `src/fragmentarium/ui/front-page/LatestTransliterationCard.tsx:51-53`, `src/fragmentarium/ui/search/FragmentariumSearchResultComponents.tsx:72-74`

Three copies of the same domain rule:

| Location | Predicate |
| --- | --- |
| `queryItemRenderReady.ts:8-11` (`hasLatestTransliterationRecord`) | `entry.type === 'Transliteration' && !entry.isHistorical` |
| `LatestTransliterationCard.tsx:51-53` | `entry.type === 'Transliteration' && !entry.isHistorical` |
| `FragmentariumSearchResultComponents.tsx:72-74` (`TransliterationRecord`) | `record.type === 'Transliteration'` — **no `isHistorical` check** |

This breaks the DRY hard gate, and the third copy is behaviourally inconsistent with the other two. `hasRenderReadyFragment(queryItem, { includeLatestRecord: true })` admits a card only when a **non-historical** transliteration record exists, but `TransliterationRecord` then renders `_(record).filter(type === 'Transliteration').first()` — which may well be a **historical** entry. On the Library "Latest Additions" list a fragment can therefore be admitted on the strength of a recent record and then display an old one.

Extract a single `getLatestTransliterationRecord(fragment)` helper next to `hasLatestTransliterationRecord` and define all three call sites in terms of it, so the admission rule and the display rule cannot drift again.

#### F2 — A malformed `museumNumber` in a summary item still takes down the whole results page

**Severity:** High · **Files:** `src/fragmentarium/infrastructure/fragmentQueryMapping.ts:66-72`, `src/fragmentarium/infrastructure/querySummaryFragment.ts:128-137`

`isQuerySummaryItemDto` is declared as `dto is QuerySummaryItemDto` but validates only `description`, `hasPhoto`, `script` and the matching-line preview. `museumNumber`, `matchingLines` and `matchCount` are never checked.

That would be a small overstatement of the guard on its own, except that `createQueryItem` dereferences `museumNumber` **before** any guard runs:

```ts
const queryItem = {
  museumNumber: museumNumberToString(dto.museumNumber),   // line 68 — runs first, unguarded
  matchingLines: dto.matchingLines,
  ...
}
if (isQuerySummaryItemDto(dto)) { ... }                    // line 73 — too late
```

`museumNumberToString` (`src/fragmentarium/domain/MuseumNumber.ts:7-11`) reads `number.suffix` with no null handling, so a missing or mis-shaped `museumNumber` throws a `TypeError` inside `.then(createQueryResult)` in the repository. That rejects the entire query promise, and `withData` renders one `ErrorAlert` in place of the whole result list.

This defeats the `UnsupportedFragmentCardSummary` → `UnavailableFragmentCard` degradation path that the rest of this PR builds so carefully: every other malformed field yields a single unavailable card, but `museumNumber` yields a blank page. Validate `museumNumber` (and ideally `matchingLines` / `matchCount`) in the guard, and move the `museumNumberToString` call after the guard so a bad item degrades to one card like every other bad item.

#### F3 — Non-line queries have no client-side next-page detection

**Severity:** High · **Files:** `src/fragmentarium/ui/search/pagination.ts:117-128`, `src/fragmentarium/ui/search/FragmentariumSearchResult.tsx:100-108`

`createPagedFragmentQuery` deliberately overfetches for line queries and not otherwise:

```ts
const needsOverfetchToDetectNextPage = isLineQuery(fragmentQuery)
limit: needsOverfetchToDetectNextPage ? pageSize + 1 : pageSize,
count: needsOverfetchToDetectNextPage ? 'exact' : 'page',
```

The consumer then computes:

```ts
const effectiveHasNextPage = lineQuery
  ? data.items.length > pageSize
  : (data.hasNextPage ?? data.items.length > pageSize)
```

For a non-line query `limit === pageSize`, so `data.items.length > pageSize` can never be true and the `??` fallback is structurally dead. If the backend does not return `hasNextPage` for a given query — omits it, or hasn't shipped it on some path — `effectiveHasNextPage` is permanently `false`, the **Next** control is permanently disabled, and users cannot page past page 1. `isPageCompletenessKnown` is also `false` in that case, so the controls stay visible while being unusable, which reads as a broken page rather than a complete one.

The line-query branch already has the right safety net. Either overfetch by one for non-line queries too and drop the backend dependency, or make the hard `hasNextPage` requirement explicit and fail loudly when it is absent instead of silently pinning users to the first page.

I could not verify which way the live backend behaves — see *Verification limitations* below.

#### F4 — `SummaryThumbnail` changes thumbnail link behaviour

**Severity:** Medium · **File:** `src/fragmentarium/ui/search/SummaryThumbnail.tsx:35-39`

```tsx
return linked ? <a href={createFragmentUrl(fragmentNumber)}>{image}</a> : image
```

The component it replaces on summary-backed results, `FragmentThumbnail` → `ThumbnailImage` (`src/common/ui/BlobImage.tsx:66`), wraps the image in `ExternalLink`, which renders `<a target="_blank" rel="noopener noreferrer">` (`src/common/ui/ExternalLink.tsx:11`).

So for summary-backed search results the thumbnail now navigates **in the same tab**, and because it is a bare `<a>` rather than a router `Link`, it triggers a **full page reload** and discards the SPA state. Two search-result cards on the same page can behave differently depending on whether the item came back as a summary or a full fragment. Use `FragmentLink` (or `ExternalLink`, to match the prior behaviour exactly) instead of the bare anchor.

#### F5 — `Reference.setDocument` clears `referenceId` but leaves `hasUnresolvedDocument` set

**Severity:** Medium · **File:** `src/bibliography/domain/Reference.ts:122-134`

```ts
setDocument(document_: BibliographyEntry): Reference {
  return produce(this, (draft: Draft<Reference>) => {
    draft.document = document_
    draft.referenceId = ''        // reset
                                  // hasUnresolvedDocument is not reset
  })
}
```

`withIdentity` sets both fields together, so they are meant to move as a pair. After `setDocument` attaches a real `BibliographyEntry`, `hasUnresolvedDocument` remains `true`, and `CompactCitation:103` keeps rendering `FallbackReferences` — the bare-id-plus-pages fallback — even though full citation metadata is now available.

Practical exposure today is small: references reaching `ReferenceForm` (`src/bibliography/ui/ReferenceForm.tsx:41`, the only production caller) come from `fragmentService.find` → `createReference`, which sets `hasUnresolvedDocument = false`. But it is a broken invariant on a domain object and will bite the first time an unresolved summary reference is fed through the editor. Reset both fields in `setDocument`.

#### F6 — `Reference.hasCitationMetadata` is dead production code

**Severity:** Low · **File:** `src/bibliography/domain/Reference.ts:54-56`

The new getter is referenced only from tests (`Reference.test.ts:62`, `createReference.test.ts:36`, `fragmentQueryMapping.bibliography.test.ts:42,88,92,107`). No production code calls it — `CompactCitation` branches on `hasUnresolvedDocument` instead. Either use it where the branching decision is made, or remove it so the domain object does not carry unused public surface that tests then lock in.

#### F7 — Two explanatory comments added, against the project style rule

**Severity:** Medium (explicit project rule) · **Files:** `src/dictionary/ui/search/FragmentLemmaLines.tsx:128`, `src/fragmentarium/ui/search/FragmentariumSearch.tsx:38`

```
src/dictionary/ui/search/FragmentLemmaLines.tsx:128:    // Keep the item-count fallback for backends that ignore count: 'exact'.
src/fragmentarium/ui/search/FragmentariumSearch.tsx:38:  // Corpus queries may still contain transport pagination fields.
```

`.github/copilot-instructions.md` states: *"Do not add comments to the code unless explicitly requested."* These are the only two non-`eslint-disable` comments added anywhere in the diff. Both facts are worth keeping — fold them into names (`itemCountFallbackTotal`, `omitTransportPaginationFields`) or into the test names that pin the behaviour, rather than into comments.

#### F8 — New files use relative imports instead of module-alias paths

**Severity:** Low (explicit project rule) · **Files:** 13 new files

`.github/copilot-instructions.md` states: *"Always use full import paths (for example `common/useObjectUrl`) instead of local relative paths like `./useObjectUrl` whenever module aliases are available."*

The new **production** module is the one that matters:

```
src/transliteration/ui/AlignmentPopover.tsx:8   import { LineGroup } from './LineGroup'
src/transliteration/ui/AlignmentPopover.tsx:9   import { Alignments } from './WordInfoAlignments'
src/transliteration/ui/AlignmentPopover.tsx:10  import LemmaInfo from './WordInfoLemmas'
src/transliteration/ui/AlignmentPopover.tsx:14  import { LineLemmasContext, useLineLemmasContext } from './LineLemmasContext'
src/transliteration/ui/AlignmentPopover.tsx:16  import { VariantAlignmentIndicator, PopoverTitle, hasLemma } from './WordInfo'
```

The remaining 12 are new test / test-support files (`FragmentView.testSupport.tsx`, `FragmentariumSearch.testSupport.tsx`, `FragmentariumSearchResult.testSupport.tsx`, `pagination.test.ts`, `withData.retry.test.tsx`, `SummaryThumbnail.test.tsx`, `AlignmentPopover.test.tsx`, `researchProjectRoutes.pagination.test.tsx`, and their siblings). It is consistent with the surrounding legacy code, which is why lint does not catch it, but the rule is stated without a test-file exemption.

For the record, the split of `WordInfo.tsx` into `AlignmentPopover.tsx` is otherwise clean and introduces no import cycle — `WordInfo.tsx` does not import `AlignmentPopover`.

#### F9 — `isLineQuery(fragmentQuery)` computed twice in the same render

**Severity:** Low · **File:** `src/fragmentarium/ui/search/FragmentariumSearchResult.tsx:101` and `:131`

```ts
const lineQuery = isLineQuery(fragmentQuery)     // line 101
...
isLineQuery(fragmentQuery) && hasLineCount       // line 131 — should be `lineQuery`
```

Harmless today, but it is the same duplication shape the DRY gate exists to prevent, and it invites the two call sites to diverge.

#### F10 — `TabController.defaultKey` can select a tab that is never rendered

**Severity:** Low (partly pre-existing) · **File:** `src/fragmentarium/ui/images/Images.tsx:45-55`

```ts
get defaultKey(): string {
  return _([this.fragment.hasPhoto && PHOTO, ...folios.map((_, i) => String(i)), hasUsableCdliTab(this.fragment) && CDLI])
    .compact().head() ?? CDLI
}
```

For a fragment with no photo, no folios and no CDLI images, every candidate is falsy and the `?? CDLI` fallback returns `'cdli'` — but this PR now gates both the CDLI nav item and the CDLI pane on `hasUsableCdliTab(fragment)`, so neither is rendered and the tab area is empty with no tab selected.

The `?? CDLI` fallback predates this PR, but the PR tightened the CDLI rendering condition (previously the pane rendered whenever `fragment.getExternalNumber('cdliNumber')` was truthy) without revisiting the fallback. Worth returning `null`/an explicit empty state instead.

Note the tightening itself is a genuine **fix**: before this PR the nav item was gated on `cdliImages.length > 0` while the pane was gated on `cdliNumber`, which was already inconsistent. And `key={fragment.number}` on `<Images>` (`src/fragmentarium/ui/fragment/CuneiformFragment.tsx:105`) correctly prevents the new `visitedTabs` lazy-mount state from leaking across fragments — that was the first thing I went looking for, and it is handled.

#### F11 — `createPeriod` swallows the parse error

**Severity:** Low · **File:** `src/fragmentarium/infrastructure/fragmentFactories.ts:31-42`

```ts
try {
  const abbreviatedPeriod = periodFromAbbreviation(periodName)
  return Periods[abbreviatedPeriod.name] ?? Periods.Uncertain
} catch {
  return Periods.Uncertain
}
```

A bare `catch` used for control flow means an unrecognised period silently becomes `Uncertain` with no signal anywhere. Given that `createSummaryItemDto` now feeds **abbreviations** (`period: fragment.script.period.abbreviation`, `src/fragmentarium/infrastructure/fragmentRepository.testSupport.ts:63-66`) while full-fragment DTOs feed **full names**, this shared code path is exactly where a backend contract drift would show up — and it would show up as every fragment silently reading "Uncertain". Prefer a non-throwing lookup, or report the unmapped value.

#### F12 — Files over the 250-line ceiling (pre-existing, not caused by this PR)

**Severity:** Informational · **Files:** `src/fragmentarium/ui/fragment/PdfExport.tsx` (776), `src/corpus/ui/ChapterViewLine.tsx` (392), `src/test-support/test-corpus-text.ts` (307)

All three were already over the limit and are touched here only by one- or two-line edits, so the PR does not push any file over. Every file this PR **adds** is comfortably under 250 lines (largest: `LatestTransliterationCard.test.tsx`, 247). Recording it because the ceiling is a stated hard gate and these will keep failing it.

#### F13 — A test file was deleted and replaced; needs explicit maintainer sign-off

**Severity:** Informational · **Files:** `src/fragmentarium/ui/image-annotation/AnnotationsView.integration.test.ts` (removed, −61), `AnnotationsView.integration.test.tsx` (added, +75), plus the removed 761-line snapshot

The project rule is *"Never remove, disable, skip, or comment out existing tests without explicit user confirmation."* On inspection this is a rewrite rather than a loss, and a good one: the `toMatchSnapshot()` assertion is replaced with explicit role-based assertions, and the full-router coverage that `AppDriver` provided moves to the new `src/router/fragmentariumRoutes.annotate.test.tsx`. The same applies to the two large deleted snapshots (`LatestTransliterations.test.tsx.snap`, 11155 lines; `FragmentariumSearch.test.tsx.snap`, 1143 lines), whose suites were rewritten around explicit assertions.

No action needed beyond an explicit "yes, approved" from the maintainer so the rule is satisfied on the record.

#### F14 — `ProjectHome` no longer renders inline search results

**Severity:** Informational · **File:** `src/research-projects/subpages/Home.tsx:47-54` (removed block)

The `{fragmentQuery && <SearchResult … />}` block is gone from the project home page. The normal flow is unaffected: `SearchForm` routes project searches to `/projects/<abbr>/search/` (`src/fragmentarium/ui/SearchForm.tsx:90-92`), which renders `SearchResult` with the new pagination props. The only loss is that a pre-existing bookmark or external link of the form `/projects/<abbr>?number=…` will now show the form with no results. Confirm that is intended.

#### F15 — Import placement in `chapter.ts`

**Severity:** Nit · **File:** `src/corpus/domain/chapter.ts:22-23`

`import { CANONICAL_ORIGIN } from 'router/domain'` sits after a blank line and directly above `export class Chapter` with no blank line between. Cosmetic only. The `CANONICAL_ORIGIN` consolidation itself (7 call sites that previously hard-coded `https://www.ebl.lmu.de`) is a good DRY win.

#### F16 — Merge state and skipped checks

**Severity:** Informational

`mergeable_state` is `blocked`. The `docker` and `docker-test` check-runs are `skipped` rather than successful. A review has been requested from `khoidt` (2026-08-17) and has not been submitted — that is the outstanding approval this PR is waiting on.

#### F17 — Affected code is short of the 100% coverage requirement, and the gaps sit on this PR's new defensive branches

**Severity:** Medium · **Measured:** full suite run with `--coverage` scoped to the PR's 45 changed production files

qlty reports 99.2% diff coverage (above its 75% threshold) and 94.0% overall (−0.1%). The project rule is stricter: *"Ensure that coverage is 100% after changes in affected code."* A local run pinpoints the gaps. The ones that matter are on **branches this PR introduces**, and several are exactly the malformed-input defences the PR is built around:

| File | Stmts | Branch | Uncovered | What is untested |
| --- | --- | --- | --- | --- |
| `src/fragmentarium/infrastructure/querySummaryFragment.ts` | 96.29 | 98.18 | `69` | `isRenderablePreviewToken`'s early `return false` for a token whose `type` is not a string — i.e. one arm of the new DTO validation |
| `src/fragmentarium/ui/images/Images.tsx` | 100 | 88.46 | `24, 65, 76-83, 175` | The whole of the new `isAvailableTab`, plus the `defaultKey` fallback in `activeKey` |
| `src/transliteration/ui/AlignmentPopover.tsx` | 93.33 | 78.57 | `96` | `AlignmentHoverTrigger`'s `onMouseEnter` / `onMouseLeave` in the newly split file |
| `src/fragmentarium/ui/search/FragmentariumSearchResultComponents.tsx` | 100 | 88.88 | `51, 108, 183` | One arm of the `usesSummaryThumbnail ? SummaryThumbnail : FragmentThumbnail` branch — directly related to F4 |
| `src/fragmentarium/infrastructure/fragmentFactories.ts` | 100 | 98.38 | `43` | The `catch` arm of `createPeriod` — the silent-`Uncertain` path from F11 |
| `src/fragmentarium/infrastructure/fragmentQueryMapping.ts` | 100 | 95.23 | `65` | The `bibliographyDocuments = {}` default-parameter branch |
| `src/bibliography/ui/CompactCitation.tsx` | 100 | 92.59 | `10, 118` | Pre-existing branches, not new |
| `src/bibliography/domain/Reference.ts` | 100 | 91.66 | `137` | `toHtml`'s `?? ''` arm — pre-existing |
| `src/router/head.tsx` | 100 | 50 | `13-14` | Default-parameter arms for `title` / `description` — pre-existing |
| `src/query/FragmentQuery.ts`, `src/query/QueryResult.ts` | 0 | 0 | — | Type-only modules with no runtime code; the 0% is an instrumentation artefact, not a real gap |

The pleasing side: `pagination.ts`, `PaginationItems.tsx`, `SummaryThumbnail.tsx`, `UnavailableFragmentCard.tsx`, `queryItemRenderReady.ts`, `LatestTransliterationCard.tsx`, `LatestTransliterations.tsx`, `fragmentCache.ts`, `ErrorAlert.tsx`, `ErrorBoundary.tsx`, `createReference.ts`, `RenderFragmentLines.tsx`, `FragmentariumSearchResult.tsx` and `router/domain.ts` are all at **100% across statements, branches, functions and lines**. The new code is well tested; it is the defensive edges that are not.

khoidt (approved):
Big change, and it lands well. Every one of the fifteen substantive findings from my last round is fixed at the root — I re-checked each against the code rather than the commit messages — and several are fixed better than I asked. F2 became a per-item fault boundary mirroring the per-card `ErrorBoundary`, so *any* malformed field costs one card instead of blanking the page; F3 dropped the backend dependency entirely by overfetching for every query rather than only line queries. Every coverage gap I tabulated last time is now closed, and the suite is 433 suites / 4336 tests green with zero console output.

Two things I want settled before this goes in. **N1** isn’t a defect I can prove — it’s a question I couldn’t answer without a backend: the client now asks for `limit=101` whenever someone picks 100 results per page, and if the API rejects that, the largest page size breaks for every search. **F13** is procedural but explicit in our rules: replacing `AnnotationsView.integration.test.ts` and deleting three snapshot files needs an explicit “yes” on the record, even though the rewrites are clearly improvements. Beyond those, **N2** and **N4** are the ones I’d most like fixed here; neither blocks.

Two things worth singling out because they’re easy to get wrong: [`fragmentCache.ts:225`](src/fragmentarium/application/fragmentCache.ts#L225) now gates prefetch caching on `hasPrefetchableFullFragment`, which requires `cardSummary === undefined` — without it a **partial** summary fragment would be cached under its museum number and then served to the fragment detail page, and that’s a subtle cache-poisoning bug that is now closed. And the three-layer degradation (DTO guard → per-item `try/catch` in the mapper → per-card `ErrorBoundary` with an `UnavailableFragmentCard` fallback) means one bad row costs one card, never the page.

### Details

**N1 — Medium (needs backend confirmation) — The client now sends `limit=101` for the 100-per-page option.** [`pagination.ts:121-131`](src/fragmentarium/ui/search/pagination.ts#L121-L131). Overfetching by one for every query is the right fix for F3 — it removes the dependency on the backend returning `hasNextPage`. The consequence is that the API `limit` is now always one greater than the page size, and `RESULT_PAGE_SIZES` is `[25, 50, 100]`, so the wire values are **26, 51 and 101**. The suite pins all three: [`fragmentariumRoutes.pagination.test.tsx:178`](src/router/fragmentariumRoutes.pagination.test.tsx#L178) (`limit: 26`) and [`:192`](src/router/fragmentariumRoutes.pagination.test.tsx#L192) (`limit: 101`). Note also that the URL parameter is *also* called `limit` and it **is** validated against `{25, 50, 100}` by `getValidatedPageSize`, so the same name means “page size” in the URL and “page size + 1” on the wire. If the backend validates `limit` against a maximum of 100 — or against an allow-list — choosing “100 results per page” returns an error for every search. I could not check this: there is no backend reachable from this container. **Fix:** confirm the API accepts `limit=101`; if it caps at 100, either lower the top page size to 99, or request `limit=pageSize` and reinstate a `hasNextPage` fallback for the largest size only.

**N2 — Medium — Summary-backed cards can never satisfy `includeLatestRecord`, so Latest Additions still refetches every card.** [`querySummaryFragment.ts:218`](src/fragmentarium/infrastructure/querySummaryFragment.ts#L218), [`LatestTransliterationCard.tsx:153`](src/fragmentarium/ui/front-page/LatestTransliterationCard.tsx#L153). `createQuerySummaryFragment` hard-codes `record: []`, and `QuerySummaryItemDto` has no record field at all, so `getLatestTransliterationRecord` returns `undefined` for *every* summary fragment and `hasRenderReadyFragment(queryItem, { includeLatestRecord: true })` is always `false`. Both Latest Additions entry points pass `includeLatestRecord: true` — `CompactFragmentCard` at `:153` and `FragmentLines` via `LatestTransliterations.tsx:79`. So the moment `/fragments/latest` starts returning **summary** items rather than full `fragments`, every card falls through to `HydratedCompactFragmentCard` → `fragmentService.find(...)`, which is exactly the N+1 this PR exists to remove. The description says Latest Additions “still depends on a backend `/fragments/latest` enhancement”, so this is known — but the blocker isn’t only that the endpoint must send prefetched fragments; it’s that a summary payload **cannot** work for this surface as the DTO is currently shaped. Worth stating explicitly so the backend contract gets a record field rather than a summary that silently defeats the optimisation. **Fix:** add the latest transliteration record (or just its date) to `QuerySummaryItemDto` and map it in `createQuerySummaryFragment`, or make `includeLatestRecord` tolerate a summary card without one.

**N3 — Medium — A white-box test reaches into `React.useState` by call index, and production code was changed to enable it.** [`withData.retry.test.tsx:85-113`](src/http/withData.retry.test.tsx#L85-L113), [`withData.tsx:33-35`](src/http/withData.tsx#L33-L35). The last test in the file, “updates the retry counter from its current value”, spies on `React.useState` and returns fakes keyed on a **call counter** (`if (useStateCall === 2) return [new Error(errorMessage), jest.fn()]; if (useStateCall === 3) return [0, setRetryCount]`). Three problems. (a) *It silently retargets*: adding, removing or reordering any `useState` in `withData` moves which hook gets faked, and the test keeps passing while asserting about different state. (b) *It asserts an implementation detail* — that the updater is `n => n + 1` — when the behaviour it stands for is already covered properly by “keeps re-querying the API on every repeated retry” immediately above it, which drives two real retries through the UI and asserts `getter` was called three times. (c) *It forced a production change*: `withData.tsx` switched `useState` from a named import to `React.useState` while leaving `useEffect` and `useRef` as named imports. I checked whether that was necessary by writing a throwaway test spying on `React.useState` against a component using the named import — it **fails**, because under this project’s Babel/Jest configuration the named binding is captured at module init, so `jest.spyOn(React, 'useState')` doesn’t intercept it. The inconsistent import style in production therefore exists solely to give this test a property it can replace. (Scratch file deleted afterwards; the tree is unmodified.) **Fix:** delete the test — its contract is already covered behaviourally — and restore the named `useState` import, so production code isn’t shaped by a test that adds nothing.

**N4 — Low — Two malformed items in one response produce duplicate React keys, and therefore console noise.** [`LatestTransliterations.tsx:48`](src/fragmentarium/ui/front-page/LatestTransliterations.tsx#L48) and [`:73`](src/fragmentarium/ui/front-page/LatestTransliterations.tsx#L73) both key on `queryItem.museumNumber`. Since F2’s fix makes `toMuseumNumberString` return `''` instead of throwing, two malformed items in the same response both key on `''` and React logs *“Encountered two children with the same key”*. That matters more than usual because we treat any console output in a test run as a defect, and because it fires in precisely the scenario this PR is built to handle gracefully. The search results list already does the right thing — [`FragmentariumSearchResult.tsx:66`](src/fragmentarium/ui/search/FragmentariumSearchResult.tsx#L66) keys on `` `${index}:${...}` ``. **Fix:** use the same composite key in both `LatestTransliterations` lists.

**N5 — Low — `'thumbnailPath' in queryItem` sniffs shape instead of using the discriminant that already exists.** [`FragmentariumSearchResultComponents.tsx:109`](src/fragmentarium/ui/search/FragmentariumSearchResultComponents.tsx#L109) and [`LatestTransliterationCard.tsx:66`](src/fragmentarium/ui/front-page/LatestTransliterationCard.tsx#L66). `QueryItem` carries an explicit discriminant — `cardSummary?: { type: 'FragmentCardSummary' | 'UnsupportedFragmentCardSummary' }` — and the sibling helper `hasUnsupportedFragmentCardSummary` uses it. These two call sites instead test for the presence of an optional property. It works today only because `createSummaryQueryItem` is the sole writer of `thumbnailPath`; the first time anything else sets it on a non-summary item, both cards silently switch thumbnail strategy. **Fix:** add `hasFragmentCardSummary(queryItem)` next to the existing predicates and use it at both sites.

**N6 — Low — `<Pagination.Item active>Page N</Pagination.Item>` is an interactive element with no action.** [`PaginationItems.tsx:110`](src/fragmentarium/ui/search/PaginationItems.tsx#L110). The current-page indicator is rendered as a `Pagination.Item`, which react-bootstrap renders as a focusable anchor: screen readers announce it as a link and keyboard users can tab to it, but activating it does nothing. `Pagination.Item` accepts `disabled`, or a plain `<li className="page-item active"><span className="page-link">` conveys “current page” without offering a dead control. The rest of this component is nicely done — `React.useId()` keeps the top and bottom instances from colliding on `htmlFor`, and `updatePageSizeSearchParam` correctly resets `paginationIndex` to 0 when the page size changes.

**N7 — Low — `SummaryThumbnail` will load an absolute URL supplied by the backend.** [`SummaryThumbnail.tsx:22-24`](src/fragmentarium/ui/search/SummaryThumbnail.tsx#L22-L24) — `const thumbnailUrl = /^https?:\/\//i.test(thumbnailPath) ? thumbnailPath : apiUrl(thumbnailPath)`. A `thumbnailPath` beginning `https://` bypasses `apiUrl` and is used verbatim as the image source, so a backend value can make every visitor’s browser issue a request to a third-party host. The backend is trusted, so this is hardening rather than a vulnerability — but the escape hatch doesn’t appear to be needed by anything, and dropping it would keep thumbnails same-origin.

**N8 — Informational — The new `ErrorBoundary` tests extend a blanket `console.error` mock.** [`ErrorBoundary.test.tsx:66`](src/common/errors/ErrorBoundary.test.tsx#L66), [`setupTests.ts:117-119`](src/setupTests.ts#L117-L119). `silenceConsoleErrors()` is `jest.spyOn(console, 'error').mockImplementation()`. Our rule says suppressing console output is “never an acceptable solution”. In fairness React’s error-boundary logging genuinely cannot be prevented at source, and the helper is pre-existing (5 files on `master`, 6 here) — the new tests simply follow the established pattern, which is the right call for consistency. Worth noting anyway: because the mock is blanket, it also swallows any *unrelated* `console.error` in the same file, including React key and `act()` warnings. Having the helper assert that what it swallowed was the expected error-boundary message would keep the gate meaningful. Not this PR’s job.

**N9 — Informational — `.gitignore` gains `.claude`.** [`.gitignore:28`](.gitignore#L28). Unrelated to the PR’s purpose, but consistent with the `.vscode/` and `.idea/` entries the repo already carries, and harmless. Recorded only so it isn’t a surprise in the diff. This is the PR’s **only** change outside `src/`.

PR #788, opened 2026-08-12
Restore corpus alignment highlight and show it for unlemmatized words

ejimsan:


src/transliteration/ui/WordInfo.tsx
Fabdulla1, line 1: So the current repository's hard blocker is that we want to refrain having files exceed 250 lines and it seems this file grew from 235 to 266 lines. We could extract the alignment related components (`AlignmentHoverTrigger`, `AlignmentInfoPopover`, and `AlignmentPopover`) into a focused sibling module

Fabdulla1 (approved):
Everything looks good. Just address the file growth as I addressed in my comment. Afterwards, it should be good to merge.

